PR #361, opened 2014-07-03
Install from db snapshots

fooman:
This pull request is aimed at speeding up the installation process (more likely for continous integration type environments).  It does so by taking a backup of the database after the initial installation and indexing has happened.

On subsequent installs with the same configuration fingerprint it will install from the database snapshot essentially saving the time it takes to

1.) upgrade from the sample demo data db version to the installed version
2.) indexing of the demo data

The backups are saved to COMPOSER_HOME/cache/_n98_magerun_download/db_snapshots


steverobbins:
:+1:

tkn98:
@fooman: You still have interest? (Just asking as there was not much traction for quite some time.)

fooman:
@tkn98 I'll close this pull request - we have used this internally in the past but will be switching to a workflow which does not require the magerun install to be fast (ie using pre-built images) so it's not likely that I'll be updating the PR.

tkn98:
@fooman: Thanks for the quick feedback.

PR #460, opened 2015-01-08
Fixed missing sprintf when creating Exception in sys:cron:run command

Eydamos:
sprintf was missing, so the two parameters for sprintf where given to the constructor of the Exception class instead resulting in the following error:
Fatal error: Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]]) in phar:///usr/local/bin/n98-magerun.phar/src/N98/Magento/Command/System/Cron/RunCommand.php on line 68


cmuench:
Thanks. Can you resend the PR for `develop` branch?

Eydamos:
New PR for develop branch is done.

Issue #466, opened 2015-01-22
Fatal Error on broken symlink on n98-magerun.yaml


tkn98:
Backtrace on error condition:

	PHP Warning:  file_get_contents(): Filename cannot be empty in phar:///usr/local/bin/n98-magerun/src/N98/Magento/Command/ConfigurationLoader.php on line 361
	PHP Stack trace:
	PHP   1. {main}() /usr/local/bin/n98-magerun:0
	PHP   2. N98\Magento\Application->run() /usr/local/bin/n98-magerun:8
	PHP   3. N98\Magento\Application->init() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Application.php:559
	PHP   4. N98\Magento\Command\ConfigurationLoader->loadStageTwo() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Application.php:595
	PHP   5. N98\Magento\Command\ConfigurationLoader->loadPluginConfig() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Command/ConfigurationLoader.php:131
	PHP   6. N98\Magento\Command\ConfigurationLoader->registerPluginConfigFile() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Command/ConfigurationLoader.php:259
	PHP   7. file_get_contents() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Command/ConfigurationLoader.php:361
	PHP Catchable fatal error:  Argument 2 passed to N98\Util\ArrayFunctions::mergeArrays() must be of the type array, null given, called in phar:///usr/local/bin/n98-magerun/src/N98/Magento/Command/ConfigurationLoader.php on line 364 and defined in phar:///usr/local/bin/n98-magerun/src/N98/Util/ArrayFunctions.php on line 20
	PHP Stack trace:
	PHP   1. {main}() /usr/local/bin/n98-magerun:0
	PHP   2. N98\Magento\Application->run() /usr/local/bin/n98-magerun:8
	PHP   3. N98\Magento\Application->init() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Application.php:559
	PHP   4. N98\Magento\Command\ConfigurationLoader->loadStageTwo() phar:///usr/local/bin/n98-magerun/src/N98/Magento/Application.php:595
[...]

PR #469, opened 2015-01-23
[BUGFIX] #466 Configuration loader fatal error

tkn98:
- Handle filesystem based config loading problems
- Give error messages that show the filename in question
- Minor WS fix


cmuench:
Thanks

PR #473, opened 2015-01-26
[TASK] Add importexporttemp table group for db:dump command

aadmathijssen:
The group is added to the @stripped group such that it not only removes the temp tables for the Dataflow module, but also for ImportExport.
This group only contains one table (importexport_importdata, which can contain both product and customer data), but has been added such that it is analogous to the handling of the dataflow temp table.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -366,8 +366,8 @@ commands:
         tables: "catalogsearch_*"
 
       - id: idx
-        description: Tables with _idx suffix
-        tables: "*_idx"
+        description: Tables with _idx suffix and index event tables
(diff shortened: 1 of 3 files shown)


aadmathijssen:
Closed due to history pollution. Will create a new clean PR.

tkn98:
Just FYI: You can rewrite the history and force-push to your branch, this will update the pull-request on github so that you won't need to create a new pull request just to have history rewritten.

aadmathijssen:
Thanks for the tip! BTW I also realised it was better to create a pull request from a feature branch, to allow more pull requests in the future.

PR #474, opened 2015-01-26
[TASK] Add index event tables to @ids table group of the db:dump command

aadmathijssen:
The index_event and index_process_event tables are needed to record the status
of dependent indexers (currently only the cataloginventory index), but are never
cleared. Excluding these tables from a database dump can significantly reduce
the size of the dump.


cmuench:
Thanks

PR #475, opened 2015-01-26
[TASK] Add importexporttemp table group for db:dump command

aadmathijssen:
The @importexporttemp group is added to the @stripped group such that it not only removes the temp tables for the Dataflow module, but also for ImportExport.
The @iimportexporttemp group only contains one table (importexport_importdata, which can contain both product and customer data), but has been added such that it is analogous to the handling of the dataflow temp tables.


cmuench:
Thanks

aadmathijssen:
You're welcome. Thanks for creating this awesome tool!

PR #478, opened 2015-01-28
[BUGFIX] Improve sys:cron:run input handling

aadmathijssen:
When sys:cron:run is called without any arguments, a dialog is presented to the user where he can choose a job by entering a number.
It was however also possible to execute a job by entering invalid input:
- if 0 is entered, the last job in the list was executed
- if a negative number was entered, the first job in the list was executed
- if non-numeric value was entered, the last job in the list was executed

This commit fixes the above issues by showing an error if incorrect input is supplied.


cmuench:
Thanks

PR #501, opened 2015-03-02
add 'assume unchanged' option for dev:log:db command, add vendor folder autoload

magicstream:


cmuench:
This is a little bit hacky. We cannot assume that every user uses git as default VCS. Please use .gitignore to exclude the file.

PR #505, opened 2015-03-06
[FEATURE] new command admin:user:toggle-active

qrz-io:
Activate/deactivate user admins.

magerun-admin-toggle-active


cmuench:
Thanks. We will merge this for next release. I hope we can find a better name as "toogle-active" :-)

qrz-io:
Thanks.

Do you have any name suggestions? I originally thought about making them 2 separate commands: `admin:user:activate` and `admin:user:deactivate`, but toggle seemed so elegant. I can go for that if you prefer it though.

cmuench:
Thanks

PR #510, opened 2015-03-11
move composer from require to require-dev

davidverholen:
Since composer is installed anyway, you only need it for development and not on the actual system, magerun is installed to.

Since @dev does only work in the root composer.json, Installation of magerun will fail if you have a minimum-stability above alpha.

Moving it to require-dev solves this, since composer still is required in your dev environment, but not anymore if you require magerun in a project

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,11 +19,11 @@
         "symfony/validator": "2.6.*",
         "n98/junit-xml": "1.0.*",
         "fzaninotto/faker": "1.4.*",
-        "composer/composer": "1.0.*@dev",
         "twig/twig": "1.*"
     },
     "require-dev": {
         "phpunit/phpunit": "4.1.*",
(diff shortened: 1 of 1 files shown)


cmuench:
@davidverholen Please note that we have really embedded composer as library.
The installert command and other parts of n98-magerun use composer to download and update data.
It's not a development requirement like phpunit.

Have a look i.e. in class: 

https://github.com/netz98/n98-magerun/blob/master/src/N98/Magento/Command/AbstractMagentoCommand.php#L5

Composer currently don't have any stable release. So we are waiting for the first non "@dev" tag.
If it is available we will add it to the root composer.json

davidverholen:
Ok, I was assuming that magerun is always installed using composer. In that case, you don't need to require composer itself.

I did not consider that magerun can also be installed as standalone, where you definitely need the requirement.

PR #514, opened 2015-03-22
Some small database connection & db:info improvements

hostep:
Hi

I was testing Magento with a db connection using sockets and noticed a couple of PHP notices like this:
`Notice: Undefined index: host`

This pull request fixes this.

While fixing this, I noticed the output of the connection strings of the `db:info` command was incorrect when using sockets, this is fixed. I also noticed the connection strings didn't contain the port when not using sockets, so this is fixed too now.

I also removed some code from `DatabaseHelper::getConnection` which was already called in `DatabaseHelper::detectDbSettings` so that wasn't necessary anymore.

I do have some extra remarks about this pull request:
- I have no experience with the JDBC connection string, so the changes are based on info I found on the web
- I hardcoded port 3306, as I see it is also hardcoded in the `InstallCommand` class, but maybe it would be a good idea to define it somewhere in a `const`?

### DIFF
--- a/src/N98/Magento/Command/Database/InfoCommand.php
+++ b/src/N98/Magento/Command/Database/InfoCommand.php
@@ -57,8 +57,9 @@ HELP;
             );
         } else {
             $pdoConnectionString = sprintf(
-                'mysql:host=%s;dbname=%s',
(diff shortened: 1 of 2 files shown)


cmuench:
Thanks

PR #526, opened 2015-04-17
Hide password when creating admin account.

### LINKED ISSUE #60: Create Admin User should hide Password
Password entry should be starred out. This may require a password-confirm prompt. Ref https://github.com/Seldaek/hidden-input for details on Symfony2 Console hidden-input field

pocallaghan:
Use askHiddenResponse to hide the password by default. To maintain
backwards compatability false isn't passed as the final argument. This
means if the terminal doesn't support hidden input plain text will be
used rather than erroring. Closes #60.


jonpday:
awesome, thx @pocallaghan

cmuench:
Thanks

PR #527, opened 2015-04-20
Fix global options swallowing arguments.

### LINKED ISSUE #521: db:info [setting] command does not work properly when --skip-root-check option is set
When --skip-root-check option is set and e.g. the command 'n98-magerun.phar --skip-root-check db:info host' is executed, the output is the same as if 'n98-magerun.phar --skip-root-check db:info' was called. So the 'host' parameter is not taken into account anymore when --skip-root-check is set.

pocallaghan:
The --skip-root-check and --skip-config options were both defined as
InputOption::VALUE_OPTIONAL. In this situation the option will be
assigned the value of an argument that is directly following it. In the
case of a command this is the command name itself, but this causes every
argument to shift along. Since these options are boolean switches, they
can be changed to InputOption::VALUE_NONE. Closes #521.

### DIFF
--- a/src/N98/Magento/Application.php
+++ b/src/N98/Magento/Application.php
@@ -156,7 +156,7 @@ class Application extends BaseApplication
         $skipExternalConfig = new InputOption(
             '--skip-config',
             '',
-            InputOption::VALUE_OPTIONAL,
+            InputOption::VALUE_NONE,
             'Do not load any custom config.'
         );
         $inputDefinition->addOption($skipExternalConfig);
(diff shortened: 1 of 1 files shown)


pocallaghan:
It's worth noting that `--root-dir` is also marked as `VALUE_OPTIONAL`, this doesn't make a lot of semantic sense, since the whole point of the option is to set a root directory. It would therefore probably make more sense as `VALUE_REQUIRED`. I didn't include this in the pull request purely because it doesn't really change the behaviour. As required or optional, it will swallow the `command` argument unless a value is specified.

cmuench:
Thanks

cmuench:
@pocallaghan I tested the changes. The solution is to use `$input->getParameterOption()` to check against a global option.

PR #528, opened 2015-04-20
Resolves #523

### LINKED ISSUE #523: Warning when using command dev:module:rewrite:conflicts
I have installed the latest n98-magerun version.
When i enter the command
```
php n98-magerun.phar dev:module:rewrite:conflicts 
```

i get all rewrites correctly as output. However, in var/log/system.log i find this error:

 > ERR (3): Warning: is_a() expects exactly 2 parameters, 3 given  in phar:///magento/n98-magerun.phar/src/N98/Magento/Command/Developer/Module/Rewrite/ConflictsCommand.php on line 152

Any suggentions or could this be caused by a third-party module? Thanks!

steverobbins:


tkn98:
@steverobbins: I reviewed your changes and I'm unsure about the fact that with this use the first operand could be string and therefore the first operand of the instanceof operation would cause a `FALSE` while it previously evaluated to `TRUE`.

https://eval.in/private/a751d74c753cfb

steverobbins:
@tkn98 I don't think the test works too well, since there aren't any rewrite conflicts in a fresh install.

steverobbins:
Rebased on `develop` since this got a revival in #523, in case it's relevant.

ktomk:
@steverobbins, thanks for care-taking. I pushed some more changes on these files, just FYI.

ktomk:
@steverobbins: Any motivation to spend some love on this older issue? We need code-changes (not only a rebase), please see comments over here: https://github.com/netz98/n98-magerun/issues/523#issuecomment-169960773

steverobbins:
@ktomk I will get to this soon.

ktomk:
@steverobbins: Aww, just seeing again that this is  a PHP versions below 5.3.9 problem. It's not even worth to take care of right now, especially as it's stale since a longer time. I don't think anyone really needs this. Perhaps just throwing an exception instead of having the implicit warning?

ktomk:
I close this for now, whenever it needs any traction, we can take care about it.

PR #546, opened 2015-05-07
added option to dump the database in xml format

### LINKED ISSUE #545: db:dump to xml format
An  improvement would be to have an option on db:dump to output on xml format. Or maybe allow to pass extra options to mysqldump from n98-magerun so we can add the --xml flag to command.

PHPUnit support xml format for database testing, it is a bit cumbersome to export tables group then import again just to have the xml in place.

redshark1802:
fixes #545


cmuench:
Thanks

PR #547, opened 2015-05-08
check error code when using db:import

### LINKED ISSUE #377: db:import does not check the error code
The db:import should check the shell's error code and display an error / stop execution.

Example:

	Optimizing sql/master.sql to temporary file
	Importing SQL dump /tmp/dumplYDrAV.sql to database foobar
	Warning: Using unique option prefix pass instead of password is deprecated and will be removed in a future release. Please use the full name instead.
	ERROR 2006 (HY000) at line 16279: MySQL server has gone away============================>                                                                                                                                     ] 32% ETA 0:00:42
	99.9MB 0:00:21 [4.54MB/s] [==============================================================================================================================================================================================>    ] 98%            

	Finished

redshark1802:
fixes #377


redshark1802:
closing this issue

implemented a solution for all exec calls in #548

PR #550, opened 2015-05-08
New Magento version with security bugfix!

rafaelstz:


cmuench:
Thanks, but the commit is already in develop branch. We will release a new n98-magerun version soon.

PR #558, opened 2015-05-15
Resolves #556: Cms block is_active toggle

### LINKED ISSUE #556: Feature Request - add ability to turn static blocks on / off
It would be very useful to be able to turn static blocks on and off similar to how your cms:banner:toggle command works.

steverobbins:


cmuench:
Thanks

PR #559, opened 2015-05-15
Resolves #46: ETA and runtimes with reindex all

### LINKED ISSUE #46: Index runtime calculator
Check index_process table and calculate the last index runtimes.

This information could also be included in the reindex actions to give an estimate on how long the index ran before.

steverobbins:


cmuench:
Thanks

PR #561, opened 2015-05-19
simplified "sales" table-group definition

mystix:


config.yaml
```diff
@@ -342,7 +342,7 @@ commands:
 
       - id: sales
         description: Sales data (orders, invoices, creditmemos etc)
-        tables: sales_order_aggregated* sales_order_tax* sales_flat* sales_recurring_* sales_refunded_* sales_payment_* enterprise_sales_* enterprise_customer_sales_* sales_bestsellers_*
+        tables: sales_* enterprise_sales_* enterprise_customer_sales_*
```
pocallaghan: This seems like it would be problematic. Tables such as sales_order_status need their values.

mystix:
Ugh, my bad... Back to the drawing board

Issue #565, opened 2015-05-26
How to run the extension on Windows?


mluzan:
We have installed your extension n98-magerun. However, when we tried to run it on Windows, it hasn't worked. Are there any condition we should meet to make it work?

redshark1802:
Hi, 

could you please provide more information:
What Windows version are you using?
What error message do you get?
What php-version do you use?

Or anything that helps us to reproduce this issue.

mluzan:
We have installed the extension according to the instructions on this page http://magerun.net/magerun-addons-by-peter-jaap-blaakmeer/ 
on Windows 7

php-version 5.6 has been used

Error message you can see if you follow this link http://screencast.com/t/cbLXqjL9QR2

And thank you for taking your time!

redshark1802:
Could you please provide the path where you put the extension, it should be located in your home directory.

Please run the following commands in your terminal and give me the output
```
php -r "echo getenv('USERPROFILE');"
php -r "echo getenv('HOME');"
php -r "echo getenv('HOMEPATH');"
```

mluzan:
So, I followed your instructions. The result http://screencast.com/t/JYekumTENc
It's here in my computer http://screencast.com/t/a9XnwIw1p and then http://screencast.com/t/bzJtt58B

redshark1802:
Please try __copying__ your ```.n98-magerun``` folder to ```n98-magerun```

mluzan:
Thank you for your help and guidance! We have succeeded.

redshark1802:
I'm glad I could help.

mageruns global foldername is inconsistent.
the folder should be named ```.n98-magerun``` on all platforms

tkn98:
+1: Folder should be named `.n98-magerun` on all platforms, Windows has no problems with dots in filenames, it's a valid filename. Good opportunity to remove conditional code.

redshark1802:
Hi @mluzan,

I've created a pullrequest to fix this issue.
Could you please rename your local ```n98-magerun```folder, checkout https://github.com/netz98/n98-magerun/pull/568 and run magerun with ```[path to git repo]/vendor/n98/magerun/bin/n98-magerun```  to test if it works.

tkn98:
I also try to test this PR on windows later.

tkn98:
We believe this flaw is fixed in the current stable version of magerun.

PR #568, opened 2015-05-27
bugfix for #565

redshark1802:
fixed inconsistent n98-magerun folder and global configuration file on windows
the global n98-magerun folder and configuration file are now prefixed with a dot (.) on all platforms

### DIFF
--- a/src/N98/Magento/Command/ConfigurationLoader.php
+++ b/src/N98/Magento/Command/ConfigurationLoader.php
@@ -119,7 +119,7 @@ class ConfigurationLoader
 
     /**
      * @param string $magentoRootFolder
-     * @param bool   $loadExternalConfig
+     * @param bool $loadExternalConfig
      * @param string $magerunStopFileFolder
      */
     public function loadStageTwo($magentoRootFolder, $loadExternalConfig = true, $magerunStopFileFolder = '')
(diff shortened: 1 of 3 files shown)


cmuench:
@redshark1802 On windows we don't like to change any path from "n98-magerun" to ".n98-magerun".
This will break some systems.
See documentation: https://github.com/netz98/n98-magerun/wiki/Modules#where-can-modules-be-placed

redshark1802:
@cmuench I totally agree that this is in fact a backwards-compatibility break.
But this issue is proof that it never worked as intended.

I still suggest doing this migration, update the docs accordingly and put it into the release notes.

ktomk:
I could fiddle with this over the weekend (found some more bugs on windows as well).

The more I look into this, the more I would consider the current behaviour on windows to be flawed. I can't find any *reason* why the directory was misnamed in the first place without the dot in front.

Current situation causes interoperability reasons, for example, to run under git-bash on windows:

    mkdir -p ~/.n98-magerun/modules/
    cd ~/.n98-magerun/modules/
    git clone git@github.com:peterjaap/magerun-addons.git

With the fix provided here, it works as documented.

I'll file a PR for the other bugs.

For this case and existing windows users, the application could be stopped if the old name is found and instructions given on how to migrate (= rename the folder-names).

ktomk:
In the next release windows will get the *nix variant with dot as fallback. That is, you can use the dotfile in windows, too. But if you have one in the homedir w/o the dot, only that one will be taken.

After release the documentation can be changed then to the new variant.

PR #573, opened 2015-05-30
Improve observer list accuracy.

pocallaghan:
If the class node isn't set in the configuration for an observer,
Magento will also check for the model node. Added support for this to
the list command.


tkn98:
Thanks for reporting and the PR.

Just a little question: What happens if none of the two are set? Are those required or optional?

For the patch itself, there is a little spacing issue and you could move the logic to create that string into a private function of it's own with two parameters: $observer and $type.

cmuench:
Thanks

PR #581, opened 2015-06-08
Updating composer.json with better dependency versions

jameshalsall:
Currently the composer.json is causing dependency hell because you're relying on specific feature versions libs instead of "at least this version, up to the next release that breaks backwards compatibility".

Symfony 2.7 components have now been released and most projects want to use these versions as they are an LTS release, but currently we can't because `n98-mageun` relies on specifically `2.6.*`.

I've also updated some other libs, and removed the `"minimum-stability": "dev"` option as you're specifying `@dev` in the composer lib dependency, which is the only one I can see is necessary to use bleeding edge (although this should be using the alpha releases).

TODO:
- [x] Update the symfony requirements to 2.7
- [ ] Fix broken tests


jameshalsall:
@tkn98 if you're happy with me changing these I will fix any deprecated usage in the Symfony components which is causing the build to fail

tkn98:
@jameshalsall: Was on holiday and offline, so answering now: Yes, switching to tilde operator looks good to me. Also removing minimum-stabilty. Two very sane changes. So yes, please if you can, go on with fixing the Symfony components. If some cause too much trouble, you can try as well with choosing some lower version or similar. Also ping me if you need some help or further feedback.

cmuench:
I  checked the CI issue. It's a deprecated warning which of the updated symfony yaml component. We can fix that with one line code. After the modification I will merge this PR.

RSully:
This would fix #646.

jameshalsall:
Apologies I've not got around to this. 

I will try and fit it in this week.

ktomk:
@jameshalsall: no worries and big thanks for spending some love on this. just ping me if there is a problem with merging or anything else that could be helpful for your work here. Also if you need support of any kind or feedback.

jameshalsall:
No worries, and thanks :).

The tests are passing now, but I think we should change the Symfony versions to be a bit more forgiving for people running 2.3 symfony components (which are still in LTS period).

If you're happy with it, I'd rather change requirements to `~2.3` which will allow people to install n98-magerun on projects where they can't update their Symfony components. This will still allow people to 2.4, 2.5, 2.6... etc.

ktomk:
@jameshalsall: I'm sorry to say it that late: I'm just seeing while I wanted to suggest you to rebase your PR on top of develop that you have PR'ed against master. We only accept PR's against the development branch. So I have to ask you to re-open this PR against develop (this requires creating a new PR on Github).

And if you could rebase it on top of develop would be great.

Regarding the versions: I don't understand the implications that well. Shouldn't it be a no-brainer unless someone uses the source version (that is not the phar version)? I might be a bit narrow-minded here as I always thought day-to-day use would be with the phar version so there shouldn't be any problem with other peoples requirements or is it? (I've just seen in another recent issue that some other user also was concerned about versions in the composer.json so a wonder a bit, maybe this is because of extensions for magerun?).

If magerun works well with ~2.3 requirements and [it's LTS](http://symfony.com/blog/symfony-2-3-0-the-first-lts-is-now-available) I see not much of a problem here generally and your suggestion sounds thoughtful to me.

jameshalsall:
No problem, I'll rebase and reopen against develop later today. 

I more frequently install magerun via composer on a per-project basis. This makes it easier when using it on a build server as we may need different versions of magerun in different projects. 

I'll see if 2.3 components work well and go with those if so

jameshalsall:
Superseded by #651

PR #586, opened 2015-06-15
[TASK] added memory_limit check

sgrund:
added memory_limit check for minimum recommended limit and Magento Gd2.php Bug on memory_limit -1

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -389,6 +389,7 @@ commands:
       php:
         - N98\Magento\Command\System\Check\PHP\ExtensionsCheck
         - N98\Magento\Command\System\Check\PHP\BytecodeCacheExtensionsCheck
+        - N98\Magento\Command\System\Check\PHP\MemorylimitCheck
       security:
         - N98\Magento\Command\System\Check\Security\LocalConfigAccessableCheck
       mysql:
@@ -425,6 +426,7 @@ commands:
         - xcache
         - Zend Optimizer
         - Zend OPcache
+      recommend-memory-limit: 512
 
   N98\Magento\Command\Media\DumpCommand:
     strip:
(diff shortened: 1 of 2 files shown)


AydinHassan:
This will return the CLI memory limit which can be completely different to the web memory limit, it might not be a good indicator?

sgrund:
You're right. My intension was to check the CLI memory limit, especially for to the magento Gd2.php bug.
I'll change the message on the consolte to CLI Memory limit, so it's clear what is checked.

tkn98:
It makes not much sense to check the CLI value at all. Often CLI is configured for unlimited (-1) and the memory is managed by the environment (soft and hard limits).

The bug in Magento is just not checking for -1, a description I could find is here: http://kostasbariotis.com/magentovarien-lib-php-memory_limit-1-bug/

The fix provided here is checking the wrong PHP configuration then.

AFAIK there is no common or smart way to obtain the PHP runtime configuration of the SAPI of which Magento is executed under.

It would be better to provide a manual checklist a user has to do on it's own against the correct configuration instead IMHO.

ktomk:
Just got an idea: Adding a Magento extension that accompanies magerun. Magerun CLI then could be able to acquire configuration information *from the same host*. The memory limit setting in this example and perhaps other settings as well it wouldn't be able to obtain from CLI directly (e.g. the user-name Magento is run under).

So far this is only a quick shoot. It's a bit convoluted.

ktomk:
When magerun is invoked (CLI), it should check on it's own for the available memory (or each command should be able to). E.g. running `sys:setup:run` with a memory limit of 128M is just not fitting (just an example). One could perhaps set the memory_limit to -1 on CLI automatically.

AydinHassan:
@ktomk I was thinking about this and was wondering that (as we know the URL of the store). During the check command, we write a PHP script to the document root which outputs all the information we need and `json_encodes` it or something. We parse the script via `file_get_contents`, `curl` whatever. Then we display the information via CLI. what do you think? This would give us information regarding the PHP configuration running the actual Magento instance, instead of just the CLI config. I think this could benefit all of the checks.

ktomk:
@AydinHassan: Like injekting a module for a single request, and then a secrect key passing with the request only magerun knows for that request?  was thinking about that as well, but it's some little work as it must be fall-safe (e.g. there is not always access to the shop-domain from each and every command-line shell due to network setup / firewall restrictions).

But yeah, that's the one Idea I had as well I could imagine if then how it could work. Maybe other commands can profit from that later on as well, but this case here would be a great test.

ktomk:
We can not check the memory-limit from the command line. Too little traction for that.

PR #594, opened 2015-06-19
fixes #593 create-all also creates setup directories

### LINKED ISSUE #593: n98-magerun.phar dev:module:create --add-all does not create setup scripts
Hi guys,

after running the `n98-magerun.phar dev:module:create --add-all` there are no setup folders are created for the module. Is it intended behavior or a bug?

luemic:


cmuench:
Thanks

PR #599, opened 2015-06-24
Excluding wishlist tables

aedmonds:
The Magento wishlist tables include a foreign key to the customer data tables. If those customer data tables are not included, the wishlist data is basically useless.

This PR excludes the wishlist tables from the database dump when the `@customers` table group is used in the `--strip` option.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -346,7 +346,7 @@ commands:
 
       - id: customers
         description: Customer data - Should not be used without @sales
-        tables: customer_address* customer_entity*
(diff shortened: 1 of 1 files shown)


tkn98:
The travis build is a bit unstable, I triggered the failed ones to run again to see this all green.

Did this trigger an error for you or was it just cumbersome to add the wishlist tables your own via command-line paramter to strip?

Just asking, considering where to start and where to stop (e.g. what about newsletter etc.).

tkn98:
Thanks.

PR #603, opened 2015-06-28
Feature/autoload restorer

ktomk:
Moved SPL based autoload restoring into utility class. Added a test-case for it.

This changes callback comparison to strict comparison.


ktomk:
green build was here: https://travis-ci.org/ktomk/n98-magerun/builds/68735322

PR #612, opened 2015-07-16
add Magento CE 1.9.2.0 (OpenMage Mirror)

### LINKED ISSUE #609: Magento 1.9.2.0 is out
Magento CE 1.9.2.0 has been released, and should be added to the installer, though it does not appear that the usual URL scheme is working any longer (for the 1.9.2.0 release).

http://www.magentocommerce.com/downloads/assets/1.9.2.0/magento-1.9.2.0.tar.gz results in a 404 error.

arosenhagen:
this PR adds Magento CE 1.9.2.0 via https://github.com/OpenMage/magento-mirror until we have a way to load the official files directly.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -220,6 +220,14 @@ commands:
         extra:
           sample-data: sample-data-1.6.1.0
 
+      - name: magento-mirror-1.9.2.0
+        version: 1.9.2.0
+        dist:
+          url: https://github.com/LokeyCoding/magento-mirror/archive/1.9.2.0.tar.gz
+          type: tar
+        extra:
+          sample-data: sample-data-1.9.1.0
+
       - name: magento-mirror-1.7.0.2
         version: 1.7.0.2
         dist:


tkn98:
Just checked in https://github.com/netz98/n98-magerun/issues/609#issuecomment-121945355 and I can see that there is not *one* 1.9.2.0 that is release by Magento officially but at least two.

Can you please make visible in your PR to which version it relates?

arosenhagen:
that's a good question :-) since the commit is 9 days old I would assume it's the first one. But there's no hint in the commit of the repo.

tkn98:
That's a bit of the problem with these unofficial sources. I assume that Magento for a reason doesn't publicly release 9.2.0.0 officially. The nearest reason is that they think they can not do this so far as they still expect changes. Not releasing publicly is speaking this clear language.

arosenhagen:
I think you're mistaken here. Magento CE 1.9.2.0 has been released officially - it's only the fact that the download can only be performed behind the login. There won't be any _changes_ to 1.9.2.0 (otherwise this would be really stupid as one cannot verify the exact version).

tkn98:
Magento CE 1.9.2.0 has not been release publicly as a stable artifact.

There additionally is not stable unified resource locator for download.

Whatever has been done - or better said proclaimed - it does not qualify as a stable public release.

arosenhagen:
ok - that might be up to your interpretation. Since there is an official release announcement (http://magento.com/blog/magento-news/magento-community-edition-192-now-available) and there is absolutely no hint that this is currently in an _unstable_ state, one must assume that this is the final version 1.9.2.0.

tkn98:
Well, don't look only at the mouth, also at the hands is the saying. We don't see a public release on a stable URI so there is obviously a problem. I don't think this is much of an interpretation at all.

And there is a hint that the software is not in a stable state: I did see changes in the version that was announced per the posting you referenced. There is no need to change a stable release. Therefore the release was *factually* not stable (it changed, if that qualifies as *unstable* depends on the stability you need. But it does not qualify as *stable*, so please mind the subtle difference).

For your PR: Please add the checksum for the download. See #610 / https://github.com/netz98/n98-magerun/commit/69cf82816244a722fa02b4610ef578098f00fa31 for examples and a bit more information (or ask in case this is not clear).

tkn98:
And please create the PR against the develop branch. Forgot to mention that.

joeyhoer:
@tkn98 @arosenhagen Magento (1) has been moving this way (content only available for authorized users) for some time, and it's possible that there will be "no [official] stable unified resource locator for download" for Magento 1 versions moving forward. It would be really great to get some feedback from the Magento team on this.

The whole download/installation process for Magneto is needlessly complex. WordPress, actually, does this quite well with their [release archive](https://wordpress.org/download/release-archive/), and [latest links](https://wordpress.org/latest.tar.gz) [including](https://wordpress.org/latest.tar.gz.md5) [checksums](https://wordpress.org/latest.tar.gz.sha1).

tkn98:
Well I don't have to finger-point over to them, it's just some little days we have checksums in magerun installer so we actually did (actually do, it's in develop so far only)  the same mistake of not providing checksums.

Next to that it really is a problem for us that we don't see a public download location for the official 1.9.2.0 package.

I think this is bad practice and in my opinion it shows that the commitment to the product is damaged on side of the vendor.

For me it's funny to see you name Wordpress as a good example. They had similar issues in the past and it's great to see they managed to change that - at least to some point. They still don't put the checksums *on* the website.

Here is another example: http://files.magerun.net/ ;)

tkn98:
@arosenhagen: I haven't heard back from you so I now created a new PR against develop with SHA checksum and an updated travis file: #614 - I've taken the URL you suggested (OpenMage). So there should be no further need for an update.

ktomk:
So far so little improvement. That 1.9.2.0 package from the github repro requires PHP 5.5.24+ to work, see https://bugs.php.net/bug.php?id=64343 - In practice this means, it currently breaks the Travis build.

There is not much information given why braindead verifications are done while it's not part of the format, but hey, please all upgrade to PHP 5.5.24+ (Mar 2015) to make use of Magento 1.9.2.0 with Magerun installer :)  (looking for more alternatives for sure).

Does anyone know how Github is packaging the tar.gz packages? Or is aware of similar problems reported?

ktomk:
Finding out more here:

> So it seems that Github uses the ***pax*** *format* for tar files, instead of the ***ustar*** *format*. For a description on tar formats, see: https://github.com/libarchive/libarchive/wiki/FormatTar [[source](http://stackoverflow.com/questions/29486029/private-repository-archivetar-gz-from-gitlab-downloaded-using-php-via-gitlab/29487303#29487303); bold/recursive by me]

And for those who are interested in a short intrduction on the history of tar formats: https://github.com/libarchive/libarchive/wiki/FormatTar

arosenhagen:
thx for the progress here. Initially I only referenced the OpenMage repo which is already mentioned for other versions. Finally I use n98-magerun in combination with [MageTestStand](https://github.com/ffuenf/MageTestStand) to automatically run tests on travis-ci - but atm there's no good way to test against 1.9.2.0 because of the 'no-official-download' issue :-/

ktomk:
@arosenhagen tests are becoming green now. I finally found a workaround to use the zip package from github, composer is not robust for Githubs .tar.gz packages due to an incompatbility of PHP with the tar format - unluckily this happened for the 1.9.2.0.tar.gz package. Anyway, this is now part of the develop version of magerun which should update automatically within the next minutes. 

To check with the test-stand, you have to try with the development version: http://files.magerun.net/n98-magerun-dev.phar

I can't yet say when the next stable release will come. I hope soon.

ktomk:
This issue has been closed via #614. The development version is available to download containing the changes. Please try it.

Issue #620, opened 2015-07-24
cache:disable does not validate code


amenk:
n98-magerun.phar ca:disable foobar

prints

`Cache foobar disabled`

Even such a cache is not existing.

Same for `cache:enable`

tkn98:
The commands `cache:enable`, `cache:disable` and `cache:clean` have been updated in the development version to report invalid cache codes/types.

The exit status is 1 in such a case.

Please try the development version and provide feedback if it solved your issue.

amenk:
seems to work

PR #625, opened 2015-08-03
Validate cache codes on cache enable, disable and clean.

convenient:
### Current functionality
```
n98-magerun.phar cache:disable eavv
Cache eavv disabled
```
### New functionality
```
n98-magerun.phar cache:disable eavv
           
  [InvalidArgumentException]  
  Invalid cache type: eavv    
                              
cache:disable [code]
```

Same for `cache:enable` and `cache:clean`.

Solves issue #620

### DIFF
--- a/src/N98/Magento/Command/AbstractMagentoCommand.php
+++ b/src/N98/Magento/Command/AbstractMagentoCommand.php
@@ -70,24 +70,6 @@ abstract class AbstractMagentoCommand extends Command
         $this->checkDeprecatedAliases($input, $output);
     }
 
-    /**
-     * @param array $codeArgument
-     * @param bool  $status
-     * @return bool
-     */
-    protected function saveCacheStatus($codeArgument, $status)
-    {
-        $cacheTypes = $this->_getCacheModel()->getTypes();
-        $enable = \Mage::app()->useCache();
-        foreach ($cacheTypes as $cacheCode => $cacheModel) {
-            if (empty($codeArgument) || in_array($cacheCode, $codeArgument)) {
-                $enable[$cacheCode] = $status ? 1 : 0;
-            }
-        }
-
-        \Mage::app()->saveUseCache($enable);
-    }
-
     private function _initWebsites()
     {
         $this->_websiteCodeMap = array();
--- a/src/N98/Magento/Command/AbstractMagentoCommand.php
+++ b/src/N98/Magento/Command/AbstractMagentoCommand.php
@@ -263,8 +263,8 @@ abstract class AbstractMagentoCommand extends Command
     /**
      * brings locally cached repository up to date if it is missing the requested tag
      *
-     * @param $package
-     * @param $targetFolder
+     * @param PackageInterface $package
+     * @param string $targetFolder
      */
     protected function checkRepository($package, $targetFolder)
     {
@@ -509,11 +509,11 @@ abstract class AbstractMagentoCommand extends Command
     }
 
     /**
-     * @param $argument
+     * @param string $argument
      * @param InputInterface $input
      * @param OutputInterface $output
-     * @param null $message
-     * @return mixed
+     * @param string $message
+     * @return string
      */
     protected function getOrAskForArgument($argument, InputInterface $input, OutputInterface $output, $message = null)
     {
--- a/src/N98/Magento/Command/Cache/AbstractCacheCommand.php
+++ b/src/N98/Magento/Command/Cache/AbstractCacheCommand.php
@@ -22,6 +22,24 @@ class AbstractCacheCommand extends AbstractMagentoCommand
     }
 
     /**
+     * @param array $codeArgument
+     * @param bool  $status
+     * @return boolean|null
+     */
+    protected function saveCacheStatus($codeArgument, $status)
+    {
+        $cacheTypes = $this->_getCacheModel()->getTypes();
+        $enable = \Mage::app()->useCache();
+        foreach ($cacheTypes as $cacheCode => $cacheModel) {
+            if (empty($codeArgument) || in_array($cacheCode, $codeArgument)) {
+                $enable[$cacheCode] = $status ? 1 : 0;
+            }
+        }
+
+        \Mage::app()->saveUseCache($enable);
+    }
+
+    /**
      * Ban cache usage before cleanup to get the latest values.
      *
      * @see https://github.com/netz98/n98-magerun/issues/483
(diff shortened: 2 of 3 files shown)


ktomk:
Thanks for providing the PR and taking care of the issue. Have you seen the minor remark on Scrutinizer? https://scrutinizer-ci.com/g/netz98/n98-magerun/inspections/fd71c28b-7c2b-4ae1-a155-37dcf424116a/issues/files/src/N98/Magento/Command/AbstractMagentoCommand.php?status=new&orderField=path&order=asc

ktomk:
And your three commits should be squashed into one as there is no reason to preserve the history of your editing for the change from what I could see. These are merely corrections of your first commit.

Please let me know if you've got issues squashing them.

convenient:
Cool. No worries. I'll give you a shout when it's all sorted.

convenient:
Would you be adverse to moving `AbstractCacheCommand::_getCacheModel` into `AbstractMagentoCommand`? 

It looks like there's some required Mage2 functionality/exception throwing that I don't want to leave out by calling `Mage::app()->getCacheInstance` directly in the `AbstractMagentoCommand`.

Seeing as there's already cache related functionality in `AbstractMagentoCommand` it makes sense to me.

tkn98:
I couldn't yet look into the code of **AbstractMagentoCommand** but if the abstract type relies on the `_getCacheModel` method already, it should be defined at least abstract in there already.

Can you link to a line where the cache related functionality already is in **AbstractMagentoCommand**?

convenient:
https://github.com/netz98/n98-magerun/blob/master/src/N98/Magento/Command/AbstractMagentoCommand.php#L80

Am I going crazy or does this line calls a method `_getCacheModel` which is only defined in `AbstractCacheCommand::_getCacheModel`.

tkn98:
You found a problem :) +1 - IMHO this is cache related so should move (if possible) into the subtype (**AbstractCacheCommand**). This should not break any tests (and hopefully no 3rd party extensions), if it breaks anything, we need to deprecate it first suggesting to use the abstract cache command instead.

What do you think?

convenient:
That seems fine. I'll refactor that all in this PR. Will update in a few hours.

tkn98:
Just seeing, **ClearCommand** uses it (via: https://github.com/netz98/n98-magerun/commit/13565bbe77a85902f7983424ebe18909a02cd38a which introduced the protected method). I wonder how this all works well. Needs review and fixing at least.

convenient:
I had a `CTRL+F` through the code base, and the only things calling  `AbstractMagentoCommand::saveCacheStatus` were `DisableCommand` and `EnableCommand` which both extend `AbstractCacheCommand`, making the method `_getCacheModel` valid.

I brought down the cache related functionality into the `AbstractCacheCommand` and everything seems to be working as normal.

tkn98:
Sorry if this sounds a bit nit-picking, but Scrutinizer often has got some good suggestions on spacing issues / code style we normally follow here: https://scrutinizer-ci.com/g/netz98/n98-magerun/inspections/b606b4ad-f66f-4f4f-9231-bc5df3a31c1f/patches - If you could review them this would be great. Thanks for squashing, this is much better for code-review.

Refactoring of the second commit looks like a good correction to me. I wonder if we should be kind and trigger a warning in the base-class so that third-party extensions that might use it have a chance to adopt. I think this is good style to add that. A simple [`trigger_error`](http://php.net/trigger_error) with a user deprecation error code and a short message that they are using the wrong base-class and instead should use the cache base class should suffice I'd say. But while writing this, I wonder how to degrade as it would not only require a warning but also a hard bail as we don't have the code there-in any longer and I don't like to have it duplicated. Maybe in the _ method to do not only the warning but also a BadMethodCall Exception. Then it's clear that the end of road has been reached. Tag it with some version in comments since when it's deprecated.

@cmuench: Can you take a look how gracefully this should be deprecated?

convenient:
No problems for the nit-pickying :+1: I'm just used to coding in PSR/2 and need to get used to working in other peoples projects.

Will have a more thorough review of your comments after work and see how it goes. Thanks.

ktomk:
Just reviewed the refactoring of pushing the two methods down to the subclass with a little distance, and there is no need to deprecate anything as this would have caused a fatal error. We don't need to deprecate fatal errors :)

convenient:
Great scott!

convenient:
No worries. I'll rewrite history a little later this evening or tomorrow :)

ktomk:
Just to make this clear: This is no requirement - and thanks for the picture :)

convenient:
Man, It's been a long day and my `git rebase-fu` is weak. What would be the best strategy for shuffling all this about so that I end up with two commits like:

1. Refactored cache specific logic out of AbstractMagentoCommand`
2. Added cache code validation to enable, disable and clean functions with totally no scrutiniser errors. 

I'm a recent migrant from SVN and am feeling it.

ktomk:
Hmm, maybe:

- checkout a new branch from your feature branch and switch back to the feature branch.
- reset it hard to develop and do the refactoring within the IDE, commit. Do the other changes - you can peek here on github or locally the newly created branch in case you're unsure to miss things and when happy do the second commmit.
- force push to origin to update the PR here.
- go to sleep or to work. after some time you get results from scrutinizer and travis so you can check with that.

I would re-do the two commits here because of the refactoring there would be a merge conflict in any case (I think). So nothing fancy.

When finally done, you can remove the newly created branch because it was just created to keep reference to the previous state for quick reference.

But as written this is not necessary. You could also just keep those two commits separated and/or squash them into one. Really not necessary.

convenient:
Thanks for that bro. I did the history refactor to get some more `git` practice. Hopefully scrutinizer shouldn't complain too much now.

convenient:
Scrutinizer is finished and is recommending I change some phpdoc in `AbstractMagentoCommand.php`. This is unrelated to the work I was dong so it can be safely ignored?

tkn98:
It's good habit to take care of the issues highlighted as new, for example: https://scrutinizer-ci.com/g/netz98/n98-magerun/inspections/bdac77f6-141e-4113-a5c2-56f3293c16cf/patches?status=new

For older code do as it pleases you. If it distracts you from your changing focus, just leave as-is. Scrutinizer also shows old issues.

convenient:
Yeah. It keep saying I've introduced phpdoc issues in that class even though all I've done was remove `saveCacheStatus` from it. I can keep going through and playing patchwork if you like, but even though they're classed as "new" they're not actually new. Don't know what is going on with scrutinizer.

tkn98:
Static code analysis is *always* the truth in it's purest form - don't question it :P I guess it's new here because it on lines of code that have changed due to the PR. I'll review this later from home.

tkn98:
@convenient Your changes have been merged into the develop branch. They are now part of the development version of magerun which is already available with unstable self-update / http://files.magerun.net/n98-magerun-dev.phar.

Thanks again for all your efforts, I have given the changes a try here already and it works like a charm!

convenient:
Cool, thanks man!.

PR #629, opened 2015-08-04
New version

rafaelstz:


ktomk:
Yes there is a new version 1.9.2.1 but magerun won't automatically configure. Compare with 1.9.2.0: #609

ktomk:
We're tracking the update in #630 now.

Issue #634, opened 2015-08-09
v1.97.3 admin:user:create hangs on CE 1.9.1.1, and EE 1.14.2.1


25thhour:
Have just tried installing `CE 1.9.1.1`,  and `EE 1.14.2.1` and noted that `admin:user:create` in `v1.97.3` is hanging in all of them — no prompt, no response, no error code.

Dropping back to v1.96.0 and `admin:user:create` *is* working on both.

Apologies but I haven't tried the other intermediate releases.

ktomk:
Thanks for reporting the issue, I can confirm the problem and a fix is on the way. Many commands within interaction might be caught by this regression, I'll update here as soon as there are any updates.

ktomk:
This has been fixed in the sources. But as there is another problem with the build system, changes aren't yet populated on files.magerun.net. The current version is available on Packagist and can be installed with composer - https://packagist.org/packages/n98/magerun .

tkn98:
The new release is out. Next fixes should come through faster again.

PR #635, opened 2015-08-09
Bugfix/issue 634 ask

ktomk:
See #634


ktomk:
Is now part of develop and master. As we have an issue with our build system, changes will not be yet visible on files.magerun.net.

PR #639, opened 2015-08-11
[TASK] Added js and css merging commands

sippsolutions:
@tkn98


sippsolutions:
@tkn98 I just fixed the mess.

ktomk:
Thanks for refreshing the PR. There is a lot of duplicate code, please take a look to reduce it.

sippsolutions:
Hi @ktomk, I don't really see any duplicate code in here :)

ktomk:
hi @sippsolutions: Both tests look nearly identical, just some parameters changed.

sippsolutions:
I copied the Test from TemplateHintsCommandTest and changed the parameters, so it should be ok (imho)

ktomk:
Yes, thank you for your contribution! :+1:  (We so desperately need more tests so we accept anything we can get :D - just kidding a bit, please see I already striked that sentence just a moment I commented)

A style thing for next time: For the commit messages we establish a new style to use active, commanding verbs in front, not the past form (here: *add* instead of *added*). Just FYI for the next time, it's not yet established as you can see with older messages.

Thanks for flying with magerun. You will see the development version updated soon. And your changes will be part of the next stable release whenever it will be released.

tkn98:
@sippsolutions: Are you okay having your full name in credits? Just asking, put it in there yesterday privately in develop, but think I should have asked you first.

sippsolutions:
@tkn98 No problem, it's fine!

PR #649, opened 2015-08-19
Adding validation to cache:flush, resolves #648

### LINKED ISSUE #648: FPC cleared - even if not
The command does not provide any error information in case FPC clearing was NOK.

steverobbins:


ktomk:
per #648

Issue #646, opened 2015-08-17
Update version of faker


RSully:
My project uses `fzaninotto/faker` version `^1.5` per `composer require`, but since magerun requires `1.4.*` I cannot use the two together (`composer require n98/magerun` fails).

ktomk:
What would be your suggestion for the requirement of `fzaninotto/faker` in Magerun? I don't have much of an insight with that library so some suggestions would be helpful. Is `~1.4` sane with that library for example? If so this would be an easy way to deal with your issue here.

RSully:
I would expect that `~1.4` is safe, but I have not looked into how you guys use it internally.

tkn98:
From top of my head it's used to generate fake customer / order data.

I'd say give it a try and create a pull request with these changes and then you can see how tests run on travis.

ktomk:
@RSully: Do you want to try a patch for this one?

jameshalsall:
The work done in #651 has fixed this issue.

RSully:
Great. @ktomk sorry it has been a few weeks since I've touched the project, but it should be fixed now if that was merged.

ktomk:
@RSully; Yes, sorry for the mass ping and no futher updates, went out of my viewport. Please give it a try and report back if it is actually fitting your needs or not. I'm with you that it should be fixed, but as you opened the issue, your feedback is greatly appreceated.

PR #651, opened 2015-08-20
Updating symfony dependencies to ~2.3 in composer.json

jameshalsall:
Currently the composer.json is causing dependency hell because you're relying on specific feature versions libs instead of "at least this version, up to the next release that breaks backwards compatibility".

Symfony 2.3 components are still in LTS, and are more suitable to depend upon. Currently we can't rely on 2.3 because `n98-mageun` relies on specifically `2.6.*`, so this is an issue when installing `n98-magerun` into a project that also relies on anything other than 2.6 Symfony components.

I've also updated some other libs, and removed the `"minimum-stability": "dev"` option as you're specifying `@dev` in the composer lib dependency, which is the only one I can see is necessary to use bleeding edge (although this should be using the alpha releases).

TODO:
- [x] Update the symfony requirements to 2.3
- [x] ~~Fix broken tests~~
- [x] Audit magerun for any potential BC breaks from using Symfony 2.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,26 +5,25 @@
     "support": {
         "issues": "https://github.com/netz98/n98-magerun/issues"
     },
-    "minimum-stability": "dev",
     "homepage": "http://netz98.github.com/n98-magerun/",
     "keywords": ["magento", "installer", "cli", "magerun"],
     "require": {
         "php": ">=5.3.3",
-        "psy/psysh": "@stable",
-        "symfony/console": "2.6.*",
-        "symfony/event-dispatcher": "2.6.*",
-        "symfony/finder": "2.6.*",
-        "symfony/yaml": "2.6.*",
-        "symfony/process": "2.6.*",
-        "symfony/validator": "2.6.*",
-        "n98/junit-xml": "1.0.*",
-        "fzaninotto/faker": "1.4.*",
+        "psy/psysh": "~0.4",
+        "symfony/console": "~2.3",
+        "symfony/event-dispatcher": "~2.3",
+        "symfony/finder": "~2.3",
+        "symfony/yaml": "~2.3",
+        "symfony/process": "~2.3",
+        "symfony/validator": "~2.3",
+        "n98/junit-xml": "~1.0",
+        "fzaninotto/faker": "~1.4",
         "composer/composer": "1.0.*@dev",
-        "twig/twig": "1.*"
+        "twig/twig": "~1.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "4.1.*",
-        "mikey179/vfsStream": "1.4.*"
+        "phpunit/phpunit": "~4.1",
+        "mikey179/vfsStream": "~1.4"
     },
     "autoload": {
         "psr-0": {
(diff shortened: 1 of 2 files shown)


ktomk:
@jameshalsall: I don't see any broken tests that need fixing  ^^ :) What we actually need to consider with your PR is that we don't break any lines with existing usage in Magerun 1 (for Magerun 2 we'll got to 2.3 LTS for the moment, I've discussed this offline today with some magerun peers, but Magerun 2 is easier to decide as the userbase is much much smaller (still)).

So to say, your feedback was great so far, I'd say the decisions done are correct as well and all we need to take care of is that we don't cut any ropes for existing users ([as you mentioned it already](https://github.com/netz98/n98-magerun/pull/581#issuecomment-132486324)).

When you review the changes, do you see anything that catches your attention? I think it's consequent to put all the symfony components on ~2.3 and it's great to see travis is green with these. I for myself have one bigger change I could test against this afterwards but would otherwise say this looks good.

So any objections from your end? Otherwise I'd say, let's try this ^^. Your change is already combined with #646 so would perhaps be a helpful signal as well.

So if you say go, I'll merge. Just wanted you to look over it a last time.

jameshalsall:
Okay cool, I'll try do an audit tomorrow - do you know what level of coverage you have from unit tests?

I will go through the changelog from Symfony and see if any of it applies to magerun and get back to you with feedback

ktomk:
Travis has coverage numbers as we do metric the percentage (at least).  You find it at the end of each build.

The lock file is committed as this is more a standalone package (see the phar file) than a library normally. I know that you mileage varies as you use the source-form. We perhaps need to play a little here.

ktomk:
@jameshalsall: We've updated and extended the build system. You should be able to see faster results on Travis now which might help you with the changes. Just pull the latest development branch and rebase your branch here to it. Faster travis = faster feedback.

jameshalsall:
I've looked through the usage of Symfony components in n98-magerun and cross-referenced it against the upgrade files in the Symfony repository. I can't see any changes that will affect the tool so I'm happy for this to be merged if you're also happy.

tkn98:
@jameshalsall  Then I'm fine to give it a try, too.

Can you - just to re-trigger the build - rebase it on top of the current develop and force push against your github repo? Then we could see it in action within the new build. Otherwise I can do this as well, just let me know.

jameshalsall:
Done :)

tkn98:
I just wonder why the composer.lock is not updated. Have you tried blanking composer.lock or something similar? Otherwise this would just change composer.json but as there is the composer.lock this would have no changes - or do I miss something obvious here?

jameshalsall:
It was updated, but github just hides the diff because it's so big, I've updated it again now so it has very latest dependency versions.

ktomk:
Oh my fault! I didn't see that (even I should know it :) )  - I was not concerned of whether the latest or not in specific. But thanks for the update. I'd say, let's then bring this - and the new build runs much faster than before that is just nice to see! Thanks for the quick turnaround time today! Let's see how it works out with the remote builds.

It's now part of the develop branch and so the unstable release. I also added another travis buld of the phar file for PHP 5.3 just to check if we can build on older systems as well.

PR #657, opened 2015-08-26
Fix database name prompt + Add space to message prefix

AydinHassan:
Looks like a regression was introduced in [ca97317f3076abb9ff06a6e582bed1aef4489caf](https://github.com/netz98/n98-magerun/commit/ca97317f3076abb9ff06a6e582bed1aef4489caf#diff-19deddad71767595667346435c7729e7R127) where `db-name` was replaced with `db-pass`. This results in the database name never being prompted for causing the command to error and exit.

I also added a space to the prompt prefix

### DIFF
--- a/src/N98/Magento/Command/LocalConfig/GenerateCommand.php
+++ b/src/N98/Magento/Command/LocalConfig/GenerateCommand.php
@@ -96,58 +96,39 @@ HELP;
     protected function askForArguments(InputInterface $input, OutputInterface $output)
     {
         $dialog = $this->getHelperSet()->get('dialog');
+        $messagePrefix = 'Please enter the ';
 
+        $arguments = array(
+            'db-host'           => array('prompt' => 'database host',        'required' => true),
+            'db-user'           => array('prompt' => 'database username',    'required' => true),
+            'db-pass'           => array('prompt' => 'database password',    'required' => false),
+            'db-name'           => array('prompt' => 'database name',        'required' => true),
+            'session-save'      => array('prompt' => 'session save',         'required' => true, 'default' => 'files'),
+            'admin-frontname'   => array('prompt' => 'admin frontname',      'required' => true, 'default' => 'admin'),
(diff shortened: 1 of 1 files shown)


AydinHassan:
hold up on this for a bit, looks like this code has other issues and can probably be cleaned up a little

tkn98:
@AydinHassan: Please share your findings. Concrete flaws should be fixed as fast as they are found, I don't see why to spare the PR, so let me know if it's fine for you to pull this in now or not.

Just FYI: I already had the one or other flaw with this (or a similar) commit, but there is no need to spare fixes only because you suspect more flaws :)

AydinHassan:
@tkn98 should be 15-20 minutes

AydinHassan:
@tkn98 

Ok Should be functional now, I've also fixed:

 * Session save prompt
 * Default value for session-save was lost in a refactor

And I've refactored it a little, any suggestions welcome

This could probably do with a test as there are lots of regressions here, I'll take a look when I get some time

tkn98:
@AydinHassan: Thanks a lot for the PR and the updates. The commit subject line is a little long, but just saying, next time you can take a look to keep it shorter and just put the details into the commit message.

I'm waiting until the build is green and then I think this is good to merge. I like how you parametrized this a bit more!

ktomk:
@AydinHassan: This is now part of the development version, I'll put your credits into the changelog soon. The unstable phar will get your changes automatically within a minute or two after the merge. I think you're right it would be nice to have a quick test for this feature but I can't give any hint how to provide the input data that would be interactive as I haven't yet written any test for interactive input. The Symfony2 console components we use perhaps has that, I know there is a command tester that at least captures output, so it might also provide input. Just in case you're looking for that. And thanks again for your contribution and for taking care!

AydinHassan:
Thanks @tkn98! I've written some tests that mock the user input before, I'll have a dig around.

PR #662, opened 2015-08-30
Unit tests for LocalConfigGenerate

AydinHassan:
As promised in #657 

* Add method to base TestCase class to get the Magento Root dir during testing
* Add tests to cover all permutations of input
* Fix bug when required argument was missing, config was still generated - now throws an exception

### DIFF
--- a/src/N98/Magento/Command/LocalConfig/GenerateCommand.php
+++ b/src/N98/Magento/Command/LocalConfig/GenerateCommand.php
@@ -43,51 +43,51 @@ HELP;
     {
         $this->detectMagento($output);
         $configFile = $this->_getLocalConfigFilename();
-        $configFileTemplate = $this->_magentoRootFolder . '/app/etc/local.xml.template';
-        if (!file_exists($configFile)) {
-            $this->writeSection($output, 'Generate Magento local.xml');
-            $this->askForArguments($input, $output);
-            if (!file_exists($configFileTemplate)) {
-                $output->writeln('<error>File ' . $this->_magentoRootFolder . '/app/etc/local.xml.template does not exist.</error>');
-                return;
-            }
+        $configFileTemplate = dirname($configFile) . '/local.xml.template';
 
-            if (!is_writable($this->_magentoRootFolder . '/app/etc')) {
-                $output->writeln('<error>Folder ' . $this->_magentoRootFolder . '/app/etc is not writeable</error>');
-                return;
-            }
-
-            $content = file_get_contents($configFileTemplate);
-            $key = $input->getArgument('encryption-key') ? $input->getArgument('encryption-key') : md5(uniqid());
-            if (is_array($key)) {
-                $key = $key[0];
-            }
-
-            $replace = array(
-                '{{date}}'               => $this->_wrapCData(date(\DateTime::RFC2822)),
-                '{{key}}'                => $this->_wrapCData($key),
-                '{{db_prefix}}'          => $this->_wrapCData(''),
-                '{{db_host}}'            => $this->_wrapCData($input->getArgument('db-host')),
-                '{{db_user}}'            => $this->_wrapCData($input->getArgument('db-user')),
-                '{{db_pass}}'            => $this->_wrapCData($input->getArgument('db-pass')),
-                '{{db_name}}'            => $this->_wrapCData($input->getArgument('db-name')),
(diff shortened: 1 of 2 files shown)


src/N98/Magento/Command/LocalConfig/GenerateCommand.php
```diff
@@ -41,52 +41,52 @@ protected function configure()
...
-                return;
-            }
+        if (file_exists($configFile)) {
+            $output->writeln(sprintf('<info>local.xml file already exists in folder "%s/app/etc"</info>', $this->_magentoRootFolder));
+            return;
+        }
 
-            $content = file_get_contents($configFileTemplate);
-            $key = $input->getArgument('encryption-key') ? $input->getArgument('encryption-key') : md5(uniqid());
-            if (is_array($key)) {
```
AydinHassan: This was introduced in e804d7d3ad8d1f4e1e6b211518651eaafed6dcb0. I removed it because the argument is not defined as an array, so an array will not be returned.

fooman: @AydinHassan see my comment here https://github.com/fooman/n98-magerun/commit/e804d7d3ad8d1f4e1e6b211518651eaafed6dcb0#comments - I never worked out why, at times, the getArgument would return an array. Adding the extra check was quicker than trying to replicate what versions and circumstances brought it up. Since it looks like you are adding test coverage, happy to see it go.

ktomk:
Thanks for this PR and the work you've put in there.

Here is something I ran over while reviewing your contribution which might profit from a generalization:

for getMagentoRoot(): In ApplicationTest::testExecute() we have the following sequence:

        $magentoRootFolder = getenv('N98_MAGERUN_TEST_MAGENTO_ROOT');
        if (empty($magentoRootFolder)) {
            $this->markTestSkipped(
                'Please specify environment variable N98_MAGERUN_TEST_MAGENTO_ROOT with path to your test ' .
                'magento installation!'
            );
        }

It's most likely worth to take over the skip-marking as normally when you call getMagentoRoot() within a test-case you're expecting to get it or otherwise skip the test.

So your new method would be an exctraction of that part. Just seeing.

ref: https://github.com/AydinHassan/n98-magerun/blob/e1e9cf5a7db85edc37bc47582371a8053a6fb174/tests/N98/Magento/ApplicationTest.php#L23

AydinHassan:
Hey @ktomk are you saying to move that logic in to `getMagentoRoot()` or copy it there? If we move it we will have to update almost all tests to call this method, right?

In any case I'm not sure the name of the method conveys it's purpose. Maybe we should just introduce a method like `requiresMagentoInstance()` which tests can call if they need a magento instance, it would skip the test if the environment is not available and as a bonus it would return the root dir. 

I think this would be nicer than the current magic of marking tests skipped as it's a bit more explicit. What do you think?

tkn98:
You raise valid points I asked myself as well. Making the name more explicit is good, I think `getTestMagentoRoot()` should make this visible. I think skipping is fine, at least it was for me and you also see the error. The alternative would be to throw some exception which would make it more explicit, but as these are test-cases auto-skip with a good message did work very well for me when I first encountered the problem (to not set the environment variable).

And regarding moving: Looks fine to me, I dunno how often it is used, for the one part I referenced, you can replace it, it's fitting for me. For other parts, I haven't reviewed, but perhaps it's not used that often? Or did I make the change at many places? ??? :)

AydinHassan:
Sorry I was getting confused, I didn't see that you added `getTestMagentoRoot` I'll update to use that instead.

tkn98:
I did already? If so, yes please use that - if it fits with your nees. Otherwise no problem to discuss / plan this further.

AydinHassan:
I already updated, build is green now. Let me know if there's anything else I need to take care of.

ktomk:
@AydinHassan: +1, very nice! I also streamlined the usage of the `getTestMagentoRoot()` method a bit on the original classes which I just saw in review for the reference I gave that I had duplicate code. ^^


While searching for the getenv of the environment variable I stumbled across **UninstallCommandTest** which is all dead code. I think we should open an issue about it to move it into a test-suite of it's own in the phpunit.xml. It perhaps is even possible to run it *afterwards* then this shouldn't hurt (according to what I read in your commit message e100705)  - at least we should be able to command that within travis to first do the default test-suite and then the one on uninstalling. What do you think?

AydinHassan:
Awesome, thanks @ktomk. Regarding `UninstallCommandTest` that's a good idea. It'll take a look at it on my lunch.

PR #664, opened 2015-08-31
[FIX] add wrapCdata test

ktomk:
the GenerateCommand has problems properly wrapping a string into a CDATA section when it comes to strings containing the word "CDATA" or "]]>".

### DIFF
--- a/changes.txt
+++ b/changes.txt
@@ -6,6 +6,7 @@ RECENT CHANGES
 1.97.6
 ======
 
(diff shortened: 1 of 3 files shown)


AydinHassan:
:+1: I considered fixing this but it was too late, I needed sleep!

tkn98:
@AydinHassan: I have a fix already but just wanted to see tests failing first. And I needed sleep, too ;) - just do your PR, I'll step behind it and will apply this little fix afterwards so there is not so much distraction for you PR.

PR #668, opened 2015-08-31
Disable coverage generation by default, enable specifically for travis

AydinHassan:
Running tests locally is horribly slow. I removed the default coverage generation and specifically enabled for travis.

This reduces tests elapsed time from over 10 minutes to 1 minute on my local machine. I think if you want coverage generation then you should specifically ask for it. In any case clover format is mostly useless to us developers. If you want coverage metrics you should just run `vendor/bin/phpunit --coverage-text`

I think we should also refactor travis to only generate coverage for maybe 1 version of php. Currently it's far too slow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,5 +65,5 @@ before_script:
   - source ./build/travis/before_script.sh
 
 script:
-  - vendor/bin/phpunit --debug --stop-on-error --stop-on-failure
+  - vendor/bin/phpunit --debug --stop-on-error --stop-on-failure --coverage-clover=clover.xml
   - php tests/check-coverage.php clover.xml ${COVERAGE}
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -27,7 +27,4 @@
             </exclude>
         </whitelist>
     </filter>
-    <logging>
-        <log type="coverage-clover" target="./clover.xml"/>
-    </logging>
 </phpunit>


tkn98:
@AydinHassan: On your local system, just place a ` phpunit.xml` file (wihtout the `.dist`) w/o code coverage. It won't be overwritten and [is excluded in the `.gitingore` file](https://github.com/netz98/n98-magerun/blob/master/.gitignore#L4).

On Travis we metric coverage to a certain percentage.

That's the current procedure - at least one I use.

tkn98:
Speeding it up is indeed a good idea. It's quite slow, but I didn't want to degrade the coverage so far. Perhaps it makes sense to use a service that is specific to code-coverage like coveralls or similar? Scrutinizer also can import the reports.

AydinHassan:
OK. I just think the default behaviour should be to not run with coverage enabled. The coverage will still be generated on travis with this change.

AydinHassan:
@tkn98 Maybe also checkout https://codecov.io/. Looks pretty cool :)

tkn98:
As travis is the default build, hard to say which way is better - I think we should make it faster. Externalizing code-coverage looks code. If you like to play around with one of the services just let me know. We then can externalize code-coverage and travis can just run trhough.

AydinHassan:
Well these services just aggregate the code coverage reports, we still have to generate them and send them to either one of the services. To speed it up we just need to decide on which combination of PHP versions and Magento Versions we should generate code-coverage for. I can't really help with that as we currently have different requirements for different combinations of Magento and PHP versions.

ktomk:
I think for code-coverage alone the PHP version is not that important. I'm setting up another CI service and try to gather coverage information with it. if this works out well we then can run the magento/php versions testssuite on travis with xdebug disabled and run coverage on another host.

ktomk:
@AydinHassan: Your idea here works out pretty well. After some hurdles I got circleci + codecov to work with the magerun testuite. travis is much faster now (from ca. 12-19 down to ca. 2-4 minutes per job) and in the meantime code-coverage runs through on circleci (ca. 25 minutes) and the coverage results then get uploaded to codecov.io. I right now have it running with my own magerun fork as a testbed, I hope to get this integrated at the beginning of the next week in the main project. This really brings things forward. Just wanted to give you a quick update, got stuck with it over the week.

AydinHassan:
Awesome, nice work @ktomk. I will look forward to the result. I'm going to try fix my other PR later today or tomorrow when I get some time.

PR #671, opened 2015-09-01
WIP - Separate test suite for Uninstall command

AydinHassan:
Currently with PHPUnit (AFAIK) there is no way to set a default test suite to run. So if a develop runs `vendor/bin/phpunit` ALL test suites will be run, and the Magento test environment will be removed. This is annoying as you cannot run the tests suite more than once. We have some options:
    
* Create a script to run tests like `runtests.php` which will just run the default test suite. (My preference)
* Update the documentation and inform to use `--testsuite` when running the tests. (I think this is a pretty annoying burden)

I'm opening this to see how it runs on travis. It *SHOULD* run the suite `n98-magerun-tests` first and then the suite `n98-magerun-magento-uninstall-tests`. I was going to separate the calls to the testsuites but I think this will affect the coverage, as the second suite will overwrite the coverage generated for the first suite. There are ways to merge coverage reports but I'm hoping we won't need to do that. In any case we shouldn't need to if the suites are ran sequentially.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -15,9 +15,13 @@
          stopOnSkipped="false"
          strict="false">
     <testsuites>
-        <testsuite name="n98-magerun Tests">
+        <testsuite name="n98-magerun-tests">
             <directory>./tests</directory>
+            <exclude>tests/N98/Magento/Command/Installer/UninstallCommandTest.php</exclude>
         </testsuite>
+      <testsuite name="n98-magerun-magento-uninstall-tests">
+          <file>./tests/N98/Magento/Command/Installer/UninstallCommandTest.php</file>
+      </testsuite>
     </testsuites>
     <filter>
         <whitelist addUncoveredFilesFromWhitelist="true">
--- a/src/N98/Magento/Command/PHPUnit/TestCase.php
+++ b/src/N98/Magento/Command/PHPUnit/TestCase.php
@@ -19,6 +19,11 @@ class TestCase extends \PHPUnit_Framework_TestCase
     private $application = null;
 
     /**
+     * @var string|null
+     */
+    private $root;
+
+    /**
      * getter for the magento root directory of the test-suite
      *
      * @see ApplicationTest::testExecute
@@ -27,6 +32,10 @@ class TestCase extends \PHPUnit_Framework_TestCase
      */
     public function getTestMagentoRoot()
     {
+        if ($this->root) {
+            return $this->root;
+        }
+
         $root = getenv('N98_MAGERUN_TEST_MAGENTO_ROOT');
         if (empty($root)) {
             $this->markTestSkipped(
@@ -35,7 +44,8 @@ class TestCase extends \PHPUnit_Framework_TestCase
             );
         }
 
-        return $root;
+        $this->root = realpath($root);
+        return $this->root;
     }
 
     /**
--- a/tests/N98/Magento/Command/Installer/UninstallCommandTest.php
+++ b/tests/N98/Magento/Command/Installer/UninstallCommandTest.php
@@ -6,59 +6,20 @@ use Symfony\Component\Console\Tester\CommandTester;
 use Symfony\Component\Console\Helper\DialogHelper;
 use Symfony\Component\Console\Helper\HelperSet;
 use N98\Magento\Command\PHPUnit\TestCase;
-use org\bovigo\vfs\vfsStream;
 
+/**
+ * Class UninstallCommandTest
+ * @package N98\Magento\Command\Installer
+ * @author  Aydin Hassan <aydin@hotmail.co.uk>
+ */
 class UninstallCommandTest extends TestCase
 {
-    /**
-     * @return string Get Magento Root
-     */
-    protected function getMageRoot()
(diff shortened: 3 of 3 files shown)


AydinHassan:
Not sure why the tests are failing,will check it out soon.

tkn98:
It probably is the case that the working directory changes, just seeing the relative file-path there. or perhaps an error with the magento version / path. Perhaps try to assert with `__DIR__` and operate relative to it. Maybe this helps.

AydinHassan:
@tkn98 seems to be working now. It is because of the Magento Root being set as a relative path. The only was I could fix that is by using realpath the first time it is asked for and cache it locally. However, I can see this being a problem if current dir is changed before anything calls that code then it will not work as expected. Maybe we should just check for then env in the bootstrap file, realpath it and re-set it? 

Also do you have any feedback regarding the points I raised in the initial description?

tkn98:
I dislike the use of realpath within code. In my eyes this is a smell. I know that there is some code in the live system that resolves any path of the magento root directory that starts with a dot ("`.`") to the current working directory (cwd). This is probably what should happen here, too. As [the code is already in](https://github.com/netz98/n98-magerun/blob/master/src/N98/Magento/Command/AbstractMagentoCommand.php#L441), it's a good candidate for code re-use (maybe in the filesystem helper class, some *resolve to current working directory* method).

Alternatively the environment variable can be set to a non-relative path. It's no problem to do that in the testsuite.

However what you brought up here is a good indicator that there is a problem in the underlying code. A relative path should work unless we change the working directory. Were you able to find out what the root-cause of this is?

Let me know if I can assist you with anything further here.

AydinHassan:
No @tkn98 I have no idea what the cause is, sorry. It works fine locally. I don't have time to debug this. It's pretty impossible to debug Travis.

Why is `realpath` a code smell? Surely we are just resolving a relative path in to it's absolute counterpart? Why is that any different from using the current directory and resolving the relative path? IMO we are just introducing more code which has the potential to contain bugs.

ktomk:
> Q: Why is realpath a code smell?

Normally whether a path is a path that has resolved all (sym)links or not must not be known by the software (most times in PHP this is the case). Only if a function explicitly needs to deal with the difference of or the resolution of links there might be cases it's actually needed.

The problem then with needlessly resolving all symlinks in a path is that the input value / path parameters are filtered / changed. Later manipulation of such paths might result in wrong results from a users perspective. Fixing these issues later can be burdensome. Allowing non-realpath()'ed path operation from the beginning (and properly checking for relative paths in the first place) is good practice in my eyes. Using realpath for defence in depth operations can be acceptable as well, however even then this often show an issue with input processing.

In the case highlighted here, it does "too much". In that case a simple

    getcwd() . '/' . $cwdRelativePath

would do it. There only is need to find out whether or not the path is relative which is the case if it does not start with "`/`" (on POSIX that would be) IIRC ([the older spec](http://pubs.opengroup.org/onlinepubs/009604499/basedefs/xbd_chap04.html)).

----

We could by definition require the environment variable to be an absolute path. This is perhaps most straight forward. More comfortable and backwards compatible is the resolution you do with realpath or when we add path resolution that is not based on realpath.

I take that in with realpath on another day we can streamline the existing code to resolve the path against the current working directory. Right now I'm too lazy to pull-up the correct POSIX docs for relative filename resolution rules.

----

> It's pretty impossible to debug Travis.

Yes, it's much work. I'll take it in as-is, I hope I can get the faster build enabled tomorrow (couldn't make it today) so that in the future Travis "debugging" at least is faster.

ktomk:
@AydinHassan: +1 for taming this beast.

AydinHassan:
Thanks @tkn98, also thanks for the information. I'll send a PR at some point to fix to `getcwd()`. It makes sense that`realpath()` is too much :+1:

ktomk:
@AydinHassan: Do you remember if the uninstall test was executed always *after* you separated the testsuites?

When I executed the testsuite locally I just realized that the test-root magento is actually removed ... . I don't remember this was the case in the past (maybe a change in phpunit?!), maybe you remember?

I consider to comment-out the uninstall test in the *phpunit.xml.dist* file and enable that test for the automated builds (Travis and Circleci).

tkn98:
In develop it's now disabled in the .xml.dist and enabled on Travis and Circleci. The forced uninstall test runs as last test.

AydinHassan:
Sorry @tkn98 I was AFK for the weekend. The test suites are ran in order so the uninstall test was ran after the normal test suite. If you don't specify which test suite to run it will run both and yes, your test-root Magento will be removed. I'm fairly certain this was always the case. 

Disabling for local sounds like an okay workaround to me :smile:

ktomk:
@AydinHassan Thanks for the feedback, this is what I thought after doing the changes  ^^ .

I just wiped a test-install accidentially by this, one I had in a PHP 7 Virtualbox for some PHP usergroup meeting back last year prior PHP 7 release, so nothing hurtful but when I ran the tests a second time and it reported me that the magent-test-root was broken I just considered to do this change ;) . Reminds once again to take care when it's about testing - and better to run some tests in isolation.

PR #679, opened 2015-09-10
Resolves #678: Show error if maintenance command fails

### LINKED ISSUE #678: Show error when maintenance command fails
It's similar to issue #412 which is about cache permissions.

When I enable maintenance mode on directory where I don't have write permissions I get false success message:

    $ n98-magerun.phar sys:maintenance --on
    Maintenance mode on

What do you think about adding error checking in `_switchOn` and `_switchOff` methods?

https://github.com/netz98/n98-magerun/blob/1a4308ed8c3ac5f5b4b94d61d3f202c0cb0d2ff9/src/N98/Magento/Command/System/MaintenanceCommand.php#L49

steverobbins:
My implementation of #678.  Let me know if it's okay.

### DIFF
--- a/src/N98/Magento/Command/System/MaintenanceCommand.php
+++ b/src/N98/Magento/Command/System/MaintenanceCommand.php
@@ -9,17 +9,26 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class MaintenanceCommand extends AbstractMagentoCommand
 {
+    const MAINTENANCE_FLAG_FILENAME = 'maintenance.flag';
+
+    /**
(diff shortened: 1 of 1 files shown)


tkn98:
@steverobbins: Thanks for providing a pull request, but there is acutally a user taking care of that flaw. I know that you take care of more here, but at least please consider to wait a little for the refactoring (fixes first) and also we can't accept PR's against master (feel free to rebase your changes on top of develop after we have the fix in and open a new PR).

ktomk:
@steverobbins: Just a quick update, original fixes are in develop now

PR #680, opened 2015-09-10
Add error checking on sys:maintenance calls

wojtekn:
Commands which enable and disable system maintenance check if touch/unlink commands returned success instead of doing that blindly.


wojtekn:
This resolves #678

tkn98:
Please only provide pull requests against the development branch, not the master branch. You can do this by rebasing your feature branch on top of current develop and create a new pull request out pf it. I close this request as the base-branch can not be changed.

PR #682, opened 2015-09-12
Enable mycli client for MySQL via db:console

chadbean:
I made a small change to the db:console command to accept the `--mycli` or `-m` switches.  If mycli (http://mycli.net/) is installed as a binary available in your path, it will be used instead of the mysql command to connect to the MySQL server when one of the above switches is set.

Using mycli is very convenient for Magento with its auto-completion features on commands and table names.


ktomk:
You can more easily use mycli with n98.magerun by aliasing it as mysql. Is there a reason why you would prefer a command-line switch over a simple alias?

Also we don't accept pull-requests aganst master but the development branch. Feel free to rebase onto develop and open a new pull-request.

PR #684, opened 2015-09-14
Enable mycli client for MySQL via db:console

chadbean:
I'm creating a new pull request based on the develop branch.

This is for a small change to the db:console command to accept the --mycli or -m switches. If mycli (http://mycli.net/) is installed as a binary available in your path, it will be used instead of the mysql command to connect to the MySQL server when one of the above switches is set.

Using mycli is very convenient for Magento with its auto-completion features on commands and table names.

It makes more sense to provide an optional switch (--mycli or -m) in n98-magerun instead of creating a custom shell alias since the impact of doing the switch is much less than changing the MySQL client via an alias. The alias may have unintentional consequences with other applications that depend on the `mysql` command to be the actual `mysql` command and not an alias to `mycli`.

### DIFF
--- a/src/N98/Magento/Command/Database/ConsoleCommand.php
+++ b/src/N98/Magento/Command/Database/ConsoleCommand.php
@@ -3,6 +3,7 @@
 namespace N98\Magento\Command\Database;
 
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class ConsoleCommand extends AbstractDatabaseCommand
@@ -12,8 +13,8 @@ class ConsoleCommand extends AbstractDatabaseCommand
         $this
             ->setName('db:console')
             ->setAliases(array('mysql-client'))
-            ->setDescription('Opens mysql client by database config from local.xml')
-        ;
+            ->addOption('use-mycli-instead-of-mysql', null, InputOption::VALUE_NONE, 'Use `mycli` as the MySQL client instead of `mysql`')
+            ->setDescription('Opens mysql client by database config from local.xml');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ktomk:
To my taste the `-m` is too short for this more or less obscure feature switch. also the long switch should be longer: `--use-mycli-instead-of-mysql` should make clear what this is about by the name alone. Otherwise the switch might not be that understandable as mycli is not well-known.

Please take in mind that mycli command is not fully compatible with the mysql command.

chadbean:
Thanks, I've implemented your suggestions (removed -m and made the long switch to be --use-mycli-instead-of-mysql).

ktomk:
@chadbean: sorry for the late feedback. I left some comments with the code and could you please squash those two commits into one? And thanks for all your efforts so far!

chadbean:
@ktomk, thanks for the feedback. I added the comma back to the $descriptorSpec array. I changed any instances of strval() to use type casting via (string). The behavior isn't really changing, but perhaps this is ideal?

tkn98:
@chadbean: you could remove strval() calls without adding (string) here because escapeshellarg()'s parameter is string. So the cast to string is implicit. So technically superfluous (you can leave it out).

Removing superfluous code normally is ideal. Less code = less maintenance.

Can you squash all the commits together so that it becomes one commit?

chadbean:
@ktomk thanks for that, these changes are now in one commit on my branch.

tkn98:
there seem to be some problems with the build. don't worry, I have an eye on it as well and restart sub-builds if needed (travis).

chadbean:
For some reason this isn't auto-filling the PW with or without the new switch. Let me look into this and let you know, thanks.

chadbean:
Thanks, I updated it to use strlen() to check for a supplied password now.

chadbean:
@tkn98 was there still an issue with the build system?

ktomk:
ups, yes there was. for some reason I can't restart it right now. I have to check this later again.

tkn98:
I can't restart this inspection, looks like a problem with Scrutinizer. I'll merge this as-is.

amenk:
@tkn98 @ktomk Multiple personalities ? ;-)

ktomk:
@amenk: Close, just one @work (employed) and the other is just my @private one.

PR #687, opened 2015-09-22
Removed duplicate assignment when detecting Magento

twslade:
Just a small syntax error with duplicate assignment statements.


tkn98:
@twslade: Thank you for your contribution, but please file your pull request against the development branch (not master as you did here).

To do that on Github you need to create a new one which is the only reason why I close that one here already. We don't accept pull requests against master.

Just ping me in case you've got any questions or need help with that.

Issue #695, opened 2015-09-25
Bash autocomplete doesn't allow path completion


bluec:
The autocomplete is a nice feature but enabling it breaks path completion.

Example:

magerun --root-dir=/[TAB][TAB]

This should start to autocomplete the path, but it doesn't do anything. Particularly annoying when using mpmd:corehacks which I realise is not a core component but it's one of the most useful magerun tools.

Can anyone confirm this behaviour? I am going to dig around some bash completion scripts and see if I can figure out how to fix it.

bluec:
Ah ok this is easier to fix than I thought:

complete -o dirnames -F _n98-magerun n98-magerun.phar
complete -o dirnames -F _n98-magerun n98-magerun
complete -o dirnames -F _n98-magerun magerun

Ref: http://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html

ktomk:
@bluec: We recently changed the bash autocomplete again, the feature you've been asking for is still available but it's perhaps a bit different now, so you probably want to try if it suits your needs. Commandline options and flags are now autocompleted, too.

PR #696, opened 2015-09-25
Bash autocomplete: Allow path completion

bluec:
Proposed fix for #695 Bash autocomplete doesn't allow path completion


ktomk:
Thanks for your PR! We do only accept pull requests against the develop branch. Please re-create it (it can't be changed), I already close the PR here. Ping me in case you need help or feedback.

PR #698, opened 2015-09-26
Bash autocomplete: Allow path completion

bluec:
Proposed fix for #695 Bash autocomplete doesn't allow path completion

Also make a single call to complete for all 3 names instead of 3 separate calls

### DIFF
--- a/autocompletion/bash/bash_complete
+++ b/autocompletion/bash/bash_complete
@@ -28,7 +28,4 @@ _n98-magerun()
     return 0
 }
 
-
-complete -F _n98-magerun n98-magerun.phar
-complete -F _n98-magerun n98-magerun
-complete -F _n98-magerun magerun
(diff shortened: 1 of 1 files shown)


bluec:
Hi Tom as requested here is a new pull request against develop branch. I also combined the three calls into a single one as you suggested.

I'm a total noob on github so finding my way around. Please let me know if I did something incorrect again. I am sure I'll get the hang of things!

tkn98:
@bluec: Have you tests if the new autocompletition with your changes works as you like it?

bluec:
@tkn98 Yes, it works well for me in my environments (all CentOS 6.7). It is a small change but a nice improvement to usability!

tkn98:
Thanks for your initiative and feedback. I've merged your changes. They will come into the master branch on the next stable release. I'm using it now from the development branch as this file isn't part of the phar build. Thanks :+1:

PR #704, opened 2015-10-07
Include custom cache-types when cleaning cache

jon-acker:
Uses getTypes from cache instance to include custom cache types as well.

By default magerun uses Mag::app->useCache() in order to retrieve the cache types to clean, but this only seems to get the core cache types. Custom cache types added by user extensions are ignored. This fixes the issue.


jon-acker:
@tkn98 thanks for merging this :+1:

tkn98:
You can delete the branch, it's a single commit so no need to keep it IMHO. And your additional report on https://github.com/inviqa/magento-symfony-container/issues/5 looks sweet!

PR #710, opened 2015-10-19
Update MySQL Version Check to handle incomplete installations

navarr:
Users may want to run a check on the codebase, in that event - if SQL hasn't been setup for a Magento project, the $dbAdapter will be a boolean value (false), and the call on (now) line 26 will result in a Fatal Error.

Checking for $dbAdapter to be truthy and setting an appropriate error if it is not, and exiting the function, allows for the script to continue on unhindered.


ktomk:
Please only create pull requests against the `develop` branch. Also please check there, it might be, that the problem is fixed in develop already (not 100% sure but your error description rings some bells).

navarr:
Yup, it's fixed in dev a different way.  My apologies.

ktomk:
No problem at all. Thank you for taking care while running over an issue! And just try with the unstable version, perhaps you're able to still provoke the error? ;)

PR #711, opened 2015-10-19
Update MySQL Engine Check to handle incomplete installations

navarr:
Users may want to run a check on the codebase, in that event - if SQL hasn't been setup for a Magento project, the $dbAdapter will be a boolean value (false), and the call on (now) line 27 will result in a Fatal Error.

Checking for $dbAdapter to be an instance of the Db_Adapter_Interface and setting an appropriate error if it is not, and exiting the function, allows for the script to continue on unhindered.

### DIFF
--- a/src/N98/Magento/Command/System/Check/MySQL/EnginesCheck.php
+++ b/src/N98/Magento/Command/System/Check/MySQL/EnginesCheck.php
@@ -15,6 +15,14 @@ class EnginesCheck implements SimpleCheck
     {
         $result = $results->createResult();
         $dbAdapter = \Mage::getModel('core/resource')->getConnection('core_write');
+        
+        if (!$dbAdapter instanceof \Varien_Db_Adapter_Interface) {
+            $result->setStatus(Result::STATUS_ERROR);
+            $result->setMessage(
+                "<error>Mysql Version: Can not check. Unable to obtain resource connection 'core_write'.</error>"
+            );
+            return;
+        }
 
         $engines = $dbAdapter->fetchAll('SHOW ENGINES');
         $innodbFound = false;


navarr:
Hey, found the same problem but elsewhere :)

ktomk:
Great! Can you provide a short description on how this is possible to reproduce? Like which command to run and I guess the XML config must be either not there or misconfigured (for the database)?

navarr:
Sure.  This error occurs when running `sys:check` on a project without a local.xml.

Error:

```PHP Fatal error:  Call to a member function fetchAll() on boolean in /home/navarr/n98-magerun/src/N98/Magento/Command/System/Check/MySQL/EnginesCheck.php on line 19
PHP Stack trace:
PHP   1. {main}() /home/navarr/n98-magerun/bin/n98-magerun:0
PHP   2. N98\Magento\Application->run() /home/navarr/n98-magerun/bin/n98-magerun:5
PHP   3. Symfony\Component\Console\Application->run() /home/navarr/n98-magerun/src/N98/Magento/Application.php:651
PHP   4. N98\Magento\Application->doRun() /home/navarr/n98-magerun/vendor/symfony/console/Application.php:126
PHP   5. Symfony\Component\Console\Application->doRun() /home/navarr/n98-magerun/src/N98/Magento/Application.php:590
PHP   6. Symfony\Component\Console\Application->doRunCommand() /home/navarr/n98-magerun/vendor/symfony/console/Application.php:195
PHP   7. N98\Magento\Command\AbstractMagentoCommand->run() /home/navarr/n98-magerun/vendor/symfony/console/Application.php:886
PHP   8. Symfony\Component\Console\Command\Command->run() /home/navarr/n98-magerun/src/N98/Magento/Command/AbstractMagentoCommand.php:432
PHP   9. N98\Magento\Command\System\CheckCommand->execute() /home/navarr/n98-magerun/vendor/symfony/console/Command/Command.php:259
PHP  10. N98\Magento\Command\System\CheckCommand->_invokeCheckClass() /home/navarr/n98-magerun/src/N98/Magento/Command/System/CheckCommand.php:72
PHP  11. N98\Magento\Command\System\Check\MySQL\EnginesCheck->check() /home/navarr/n98-magerun/src/N98/Magento/Command/System/CheckCommand.php:94
```

ktomk:
Yes, getConnection returns `NULL`  in this case IIRC. Were you able to search other places where this is called? It's probably worth I just check this as well while we're at it... .

navarr:
`N98\Util\Console\Helper\DatabaseHelper` and anywhere it's called does not do error checking, though that's normally for MySQL specific commands.

A quick search found these instances of where the database connection is used but there is no error handling in the event it does not exist.

* `N98\Magento\command\System\Setup\RemoveCommand::removeSetupResource`
* `N98\Magento\Command\Developer\Translate\ExportCommand::execute`
* `N98\Magento\Command\Developer\Setup\Script\AttributeCommand::execute`
* `N98\Magento\Command\Developer\Code\Model\MethodCommand::initTableColumns`
* `N98\Magento\Command\Database\Maintain\CheckTablesCommand::_queryAlterTable`

ktomk:
Hmm, I dunno if I have an update of that helper over at the other box here. I think if it throws a runtime exception, this will be automatically handled. So technically the helper should provide the connection and throw the exception if not available. Then all other code won't need to care about it I think.

However this requires that the other places to use the helper which don't use it now (e.g. the place you reported here) should be changed to use the helper instead.

What do you think?

ktomk:
@navarr: FYI, I've refreshed a branch which contains related commits, here is a specific commit regarding the error handling of the XML file processing: https://github.com/ktomk/n98-magerun/commit/1edb42b35265ab908abc418227d79eeaa73a8b01 - the whole feature/interop branch I have there contains more database helper related changes which should stabilize the use.

navarr:
:+1:

ktomk:
@navarr: Thanks again for the report, even reply now is late.

Fixes were incorporated in database helper already, for this one place (and another one), fix is in bd8745f1705de3e74ec535411660c996b8c64108 now. Will be part of the next stable release.

PR #730, opened 2015-11-10
Fix #729: Ensure YAML::parse receives file contents instead of path.

### LINKED ISSUE #729: ConfigurationLoader->loadDistConfig should not be passing file path to Yaml::parse
The `ConfigurationLoader` class' `loadDistConfig` method passes the `config.yaml` file path instead of its file contents. From Symfony 3 onward, that ability will be removed:

```
Deprecated: The ability to pass file names to the Symfony\Component\Yaml\Yaml::parse method is deprecated since version 2.2 and will be removed in 3.0. Pass the YAML contents of the file instead. in /vendor/symfony/yaml/Yaml.php on line 55
```

The method:
```php
    protected function loadDistConfig($initConfig)
    {
        if ($this->_distConfig == null) {
            $this->_distConfig = Yaml::parse(__DIR__ . '/../../../../config.yaml');

            if (OutputInterface::VERBOSITY_DEBUG <= $this->_output->getVerbosity()) {
                $this->_output->writeln('<debug>Load dist config</debug>');
            }
        }

        $config = ArrayFunctions::mergeArrays($this->_distConfig, $initConfig);

        return $config;
    }
```

danemacmillan:


ktomk:
I took over your change into develop (and extended it a bit). Please next time do a pull request against the develop branch. We don't accept PRs against master.

PR #733, opened 2015-11-13
FIX: show all available extension versions (alpha, beta, stable) with…

EmPeWe:
show all available extension versions (alpha, beta, stable) with "extenions:search"

### DIFF
--- a/src/N98/Magento/Command/MagentoConnect/AbstractConnectCommand.php
+++ b/src/N98/Magento/Command/MagentoConnect/AbstractConnectCommand.php
@@ -63,7 +63,9 @@ abstract class AbstractConnectCommand extends AbstractMagentoCommand
     protected function matchConnectLine($line)
     {
         $matches = array();
-        preg_match('/([a-zA-Z0-9-_]+):\s([0-9.]+)\s([a-z]+)/', $line, $matches);
+        // expected format "Package: Version Stability[,Version Stability][,Version Stability]"
+        $pattern = '/([a-zA-Z0-9-_]+):\s([0-9.]+)\s([a-z]+),?([0-9.]+)?\s?([a-z]+)?,?([0-9.]+)?\s?([a-z]+)?/';
+        preg_match($pattern, $line, $matches);
         return $matches;
     }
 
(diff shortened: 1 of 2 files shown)


ktomk:
Please make this one commit (squash the second one into the first).

And if I see it right, this is limited to a maximum of three version identifiers per entry.

Even already an improvement over the current situation that only reflected one, the implementation is short-sighted as it's still limited. This time to a maximum of three versions (instead of one version).

I think it's more effective with a loop. Let me know if you need help.

EmPeWe:
I don't know how to squash but yes, I could try.

mage list-available will never show more than 3 versions, because only 3 preferred states exists: alpha, beta and stable. The mage command will never list more than 1 version per state, but the latest release for each (if existing). A loop may be elegant, but in this case it will be overcrafted and I prefer clean, readable und understandable code over elegance.
Think about another developer will read this code. A loop would tell 1 to infinite results. My proposal is telling 1 to 3 results, which is the maximum preferred states available in magento. There will never be a fourth result.

Do we have a deal?

ktomk:
@EmPeWe Thanks for your detailed feedback, I was not aware that there aren't more than three to be expected. In that case your solution for sure is sufficient. And there is no need for a deal, instead please excuse my ignorance I just see you've left a comment within the code.

How to squash: You can find a description of it here: 

* [squashing commits with rebase (by Nick Quaranto)](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

After you squashed, do a force push to the repository on github and it will update the pull request here.

I can pull the changes in then. Thanks a lot!

EmPeWe:
Thx for the explanation. I've just squashed and pushed to master.

Thank you for your support.

ktomk:
Okay, then I'll do a manual merge, because I oversaw you've done it against master and we only take against develop. But you've done so much already, I'll take care, no worries. Will do it in 1-2h.

EmPeWe:
Damn it. I'm sorry that I pushed to master.
I hope, that you won't have excessive extra work to do. Next time, I'll try to make it better.

tkn98:
My fault, I normally tell it *at the beginning* not at the end. Checks are green now, travis got stuck. I'll take care now.

tkn98:
I've taken your changes over into the development branch.

See https://github.com/netz98/n98-magerun/commit/f8a7da593951ad9a95b4528746aece3e5946e07b

You should be able to try it out in the unstable release channel.

Issue #727, opened 2015-11-09
Cron Job List command sys:cron:list does not merges system config path values


atishgoswami:
The `sys:cron:list` command does not consider the path values saved in the `core_config_data` table.

Example:
In the config.xml we have the following:
```xml
<crontab>
        <jobs>
            <foo_bar_cron>
                <run>
                    <model>foo_bar/observer::runFooJob</model>
                </run>
            </foo_bar_cron>
        </jobs>
</crontab>
```
and in `core_config_data` table we have:

Path  = `crontab/jobs/foo_bar_cron/schedule/cron_expr`
Value = `0 0 * * 1`

When we run `php n98-magerun.phar sys:cron:list` output 
does not show the scheduled time

```
+------------------------------------------------+------+----+----+----+----+
| Job                                            | m    | h  | D  | M  | WD |
+------------------------------------------------+------+----+----+----+----+
| foo_bar_cron                                   |      |    |    |    |    |
+------------------------------------------------+------+----+----+----+----+

```

ktomk:
Is this also the case after flushing the cache?

atishgoswami:
@ktomk I don't have cache enabled in my development environment.

tkn98:
Can you just for clarity add as well the expected result? I guess it's 

    +------------------------------------------------+------+----+----+----+----+
    | Job                                            | m    | h  | D  | M  | WD |
    +------------------------------------------------+------+----+----+----+----+
    | foo_bar_cron                                   |  0   | 0  | *   | *  | 1   |
    +------------------------------------------------+------+----+----+----+----+

so that magento merges XML + config paths, correct?

atishgoswami:
@tkn98 Exactly !!

tkn98:
The fix is now part of the development version. Please update to the non-stable variant and report back if it fixed/didn't fix your issue. Thanks!

atishgoswami:
@tkn98 Sure will test it out. Thanks for the update

ktomk:
No feedback provided, I assume in good faith that it is fixed.

atishgoswami:
Hello,

Sorry for the late reply got a it little distracted with other things.

Yes the fix seems to be working !!!

Thanks @robbieaverill and @tkn98

ktomk:
[...]

PR #739, opened 2015-12-18
[#727] Merge DB cron expressions from the database into the XML

robbieaverill:
This pull request addresses issue #727 where cron expressions in the `core_config_data` table were not being displayed when running `sys:cron:list`.

In accordance with the [Magento core cron model](https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Cron/Model/Observer.php#L86-L98), we should load both and merge them together.

Matching core functionality, the database entries take priority over XML entries.


robbieaverill:
Fixing unit tests...

tkn98:
@robbieaverill: Nice fix! Please see my small comment, I'll fix this when I edit the changelog.

Issue #741, opened 2015-12-29
There are no commands defined in the "customer" namespace.


MrRoyce:
I installed, updated and tested magerun, and I can run some of the commands such as 'sys:check', etc.  But when I try to get customers with 'customer:list', I get the error message 

Invalid Argument Extension,  There are no commands defined in the "customer" namespace.

Any idea how to resolve this?

ktomk:
That sounds like as-if there would be no customer:* command at all. When you call the list command, is it shown in the list of all commands?

MrRoyce:
Nope, no customer commands at all on the server.  The commands go from config to db.

It works fine on my localhost.  Both are using 1.97.7

ktomk:
I tried with 1.97.7 and 1.97.8. Please run the selfupdate command to upgrade to the current version first. I'm not sure if that solves your problem but it's best if you're checking against the latest version. If there is any reason you can't upgrade, just let me know as well.

MrRoyce:
I ran the selfupdate which says its updating to 1.97.8, but the version command still shows 1.97.7

ktomk:
@MrRoyce: I just verified what you comment and can confirm there is a problem with the download of the latest version. But nevertheless, this should not trigger your original issue, so let's for the moment ignore it (created #742).

In [your previous comment](https://github.com/netz98/n98-magerun/issues/741#issuecomment-167933987) you wrote that 

> Nope, no customer commands at all on the server (with PHP 5.3.24 - shared hosting). 

That is a clear sign that on your system the PHP function [`exec`](http://php.net/exec) is not available. Unless that *exec* function  is available no customer commands are enabled.

I hope this clarifies your issue.

Can you share which hoster you're using and can you confirm that the exec function is disabled on that shared hosting?

MrRoyce:
You are correct, exec is not available on the server.  I'd rather not mention the hosting company. I inherited this project that had been hastily switched to them by the prior developer.  I had a long discussion with them on Twitter and email with their Community Developer about why they would default to PHP 5.3. for shared hosting but I guess something like 'exec' being disabled is reasonable. 

Thankfully this is just a temp server that can be scrapped.  Everything works fine on my local, 

BTW I searched a bit, but didn't see any info on exec being a requirement for Magento?  Or is it Magerun that needs it?  Is that and any other system requirements documented anywhere?

I'll close this ticket.  Thanks for your help.

ktomk:
[...]

PR #743, opened 2015-12-31
[FIX] #741 remove customer exec dependency

ktomk:
remove isEnabled() check from the abstract customer command.

by it the abstract customer command was checking for the exec() function
but all of the customer commands do not need it.

the flaw cam to light in issue #741.

the check has been introduced in error in 5555a78 .

the fix is to remove the check and to re-add it in the correct places:

- Database\ImportCommand
- Database\QueryCommand
- Installer\InstallCommand
- OpenBrowserCommand
- ScriptCommand

As the check was originally with some other commands, in Exec the
allowance-check hast been additionally added into Exec::run() to give a
more saying message in a RuntimeException instead of a fatal error.


ktomk:
Meged in 21335469886cd5d9d1a21f37e5ec0af786708997, closes #471

PR #744, opened 2016-01-04
[FIX] Install command fails if usting custom PHP version

jsiefer:
Using a custom PHP version, install command fails as it uses the default OS version.
Works fine on OSX, not sure about Win


ktomk:
Hi @jsiefer, please file the pull-request against the development branch (you need to create it new, that's why I closed it already). Also please check if not the [`PHP_BINARY`](http://php.net/manual/en/reserved.constants.php) reserved constant is more fitting:

>  PHP_BINARY (string)
    Specifies the PHP binary path during script execution. Available since PHP 5.4.

Thanks!

ktomk:
new PR is:  #751

PR #748, opened 2016-01-07
Return non-zero status when module rewrite conflict is positive

steverobbins:
This makes the `dev:module:rewrite:conflicts` command return a non-zero exit code if a rewrite conflict is detected.  This is useful for integration testing, so that a build will be marked as failed when a conflict is introduced.


ktomk:
Very useful, I remember this was reported earlier, but haven't found any issue. This is now part of the unstable development version and will be in the next stable release. Thank you!

PR #751, opened 2016-01-08
[FIX] install cmd fails using custom php binary

jsiefer:
Use PHP_BINARY constant if available to run the magento install command using the same PHP binary as magerun.
Also fix minor issue for running phpunit on osx/darwin


ktomk:
PR looks fine, great to see the test-case fixed for Mac OSX. I only found this one comment that is imprecise about the PHP version.

jsiefer:
thanks, missed that one, but I will close this PR and do it again against the dev-branch.

ktomk:
New PR is #752

PR #762, opened 2016-02-02
Adding a prompt when unlocking all admin users.

brobie:
There have been numerous time that I run this command, expecting a prompt for a username to be entered, and it simply unlocks all users. I have added a prompt to make sure that the intention is to unlock everyone.

### DIFF
--- a/src/N98/Magento/Command/Admin/User/UnlockCommand.php
+++ b/src/N98/Magento/Command/Admin/User/UnlockCommand.php
@@ -49,12 +49,15 @@ class UnlockCommand extends AbstractAdminUserCommand
                 $output->writeln('<info><comment>' . $username . '</comment> unlocked</info>');
                 return;
             }
-            \Mage::getResourceModel('enterprise_pci/admin_user')->unlock(
-                \Mage::getModel('admin/user')
-                    ->getCollection()
-                    ->getAllIds()
-            );
-            $output->writeln('<info><comment>All admins</comment> unlocked</info>');
+            $dialog = $this->getHelperSet()->get('dialog');
+            $shouldUnlockAll = $dialog->askConfirmation($output, '<question>Really unlock all users?</question> <comment>[n]</comment>: ', false);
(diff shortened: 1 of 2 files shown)


ktomk:
@brobie: Thanks for your updated PR. I've left some comments on the commit but those are merely little details that could be touched but I don't think there is anything required to get it to work. Perhaps if you review my comments you find a place you want to touch  then please do.

The biggest problem I had when reviewing the PR was that the test you provide does only work for an enterprise version installed. As we run the tests and the coverage with CE versions only (travis/circleci/codecov) the tests are skipped.

I tried to mock this over the last seven days again and again but was not successful with it. Have you tried to test it with a mock so that coverage works with a CE version or was it that you just didn't really consider it as the command is disabled then anyway?

Please give feedback but if not I'll merge then as-is.

brobie:
Because the command was disabled for CE, I figured that I would skip it
since you are only testing for CE.  I mainly wrote the test to prove that
it works for those that want to run it locally against an EE version.  I'll
check out the other comments.

ktomk:
The test with the skipping is well done for CE environments. And I can imagine it works for EE. Please keep it that way.

For the other comments, please do check them out, but as written, I do not think further changes are important. Perhaps check the interaction logic you might consider some kind of change useful.

Just while writing I have the following idea: What if it's asked to unlock:

1. none (cancel)
2. all (really unlock all users)
3. Admin User A (a specific user)
4. Admin User B (the second specific user)
5. ... (all the other specific users, one after another)

would that make sense? Then while entering the command, all options would be asked. But anyway, that would be additional material and then I would need really a EE version to test here (which I should make real tomorrow anyway

ktomk:
After #808 / #810 this PR has become stale. The patch on the command might need a little change, the tests in themselves should still do their job.

ktomk:
Fixing in a new PR, this feature will ship with the next stable release.

ktomk:
@brobie: Thanks!

PR #763, opened 2016-02-03
Sanitize module list properties

bennoislost:
if a module definition property (version,codepool) contains a whitespace character the output table display will be broken and the filters wont apply correctly.

By removing whitespace when collecting the module properties the output table will render correctly (even with a whitespace character defined in the modules definition file) and filters will still correctly apply.


ktomk:
@bennoislost: Please do a pull request against the development branch, you made this one against master (you have to create it new, just with a different base).

PR #764, opened 2016-02-03
Sanitize module properties

bennoislost:
if a module definition property (version,codepool) contains a whitespace character the output table display will be broken and the filters wont apply correctly.

By removing whitespace when collecting the module properties the output table will render correctly (even with a whitespace character defined in the modules definition file) and filters will still correctly apply.


ktomk:
@bennoislost: Thank you for creating (and updating :)) this pull request. The changes are now part of the development version and will become part of the next stable release.

PR #766, opened 2016-02-06
add official version of magento-ce-1.9.2.0 - magento-ce-1.9.2.3 / switch to zip format

arosenhagen:
as for versions < 1.8.0.0 only zips are available through the download API we might switch generally to zips (otherwise checksum verification will fail in 3rd party implementations or we have to deal with superfluous conditionals)
also add the official sources for 1.9.2.0 - 1.9.2.3 (these still need the magedownload-cli approach)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -170,6 +170,15 @@ commands:
 
   N98\Magento\Command\Installer\InstallCommand:
     magento-packages:
+      - name: magento-ce-1.9.2.4
+        version: 1.9.2.4
+        dist:
+          url: http://www.magentocommerce.com/downloads/assets/1.9.2.1/magento-1.9.2.3.zip
+          type: zip
+          shasum: 219fe2b88068cf73953b90ce01186cb1f19affea
+        extra:
+          sample-data: sample-data-1.9.1.0
+
       - name: magento-ce-1.9.2.3
         version: 1.9.2.3
         dist:
--- a/config.yaml
+++ b/config.yaml
@@ -170,66 +170,102 @@ commands:
 
   N98\Magento\Command\Installer\InstallCommand:
     magento-packages:
+      - name: magento-ce-1.9.2.3
+        version: 1.9.2.3
+        dist:
+          url: http://www.magentocommerce.com/downloads/assets/1.9.2.1/magento-1.9.2.3.zip
+          type: zip
+          shasum: c211c1d31468dfbc45decd7d00266281fdc97b20
+        extra:
+          sample-data: sample-data-1.9.1.0
+
+      - name: magento-ce-1.9.2.2
+        version: 1.9.2.2
+        dist:
+          url: http://www.magentocommerce.com/downloads/assets/1.9.2.1/magento-1.9.2.2.zip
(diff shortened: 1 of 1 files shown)


ktomk:
I'm fine with more consistency, but for review it would require me to obtain those zip files, also the older ones. Do you have a source for these?

And for the magedownload-cli approach, can you provide a *quick* reference?

arosenhagen:
sorry for the delay. basically it'll be great to incorporate @steverobbins https://github.com/steverobbins/magedownload-cli into n98-mageruns core so we can continue to download from the official sources. I currently use it in https://github.com/ffuenf/MageTestStand/blob/master/install.sh#L55 outside of n98-magerun for automated tests on travis. The install command further down recognizes the previously downloaded zip (load from cache if download is already on the filesystem).

tkn98:
@arosenhagen: Quick question: Aren't tgz pacakges available? AFAIK they should work (and be smaller than zip), problem with tgz from Github was because they use a tar extension not supported by the PHP Phar extension (which is used for unpacking).

arosenhagen:
unfortunately old versions are really only available as zip (I would also prefer tar.gz):
```
magento-1.7.0.2.zip  Community Edition - Full
magento-1.7.0.1.zip  Community Edition - Full
magento-1.7.0.0.zip  Community Edition - Full
magento-1.6.2.0.zip  Community Edition - Full
magento-1.6.1.0.zip  Community Edition - Full
magento-1.6.0.0.zip  Community Edition - Full
magento-1.5.1.0.zip  Community Edition - Full
magento-1.5.0.1.zip  Community Edition - Full
magento-1.4.2.0.zip  Community Edition - Full
magento-1.4.1.1.zip  Community Edition - Full
magento-1.4.1.0.zip  Community Edition - Full
magento-1.4.0.1.zip  Community Edition - Full
magento-1.4.0.0.zip  Community Edition - Full
```
I know these are very old and by no means recommended anymore.

ktomk:
Puh, I just checked one of the zips in the PR and I can't download them, it's the standard 404. So I'd say this makes no sense. We could add downloads from a mirror. If there is a script which can create a mirror by downloading automated with ID/token from magento could be helpful. But that's most likely out of scope from Magerun. Perhaps *if* it's possible to add a driver for the Mageno API, this could work better. It's possible to store the crendetials within a user-config (and/or provide it with a commandline parameter like `<id>:<token>` format or similar.

ktomk:
@arosenhagen: I'm going through older requests. Can you help me and say what the benefit would be if the download does not work?

arosenhagen:
@ktomk well the benefit would be, that all of the official zip sources would work when using the magerun version of https://github.com/steverobbins/magedownload-cli.
Currently I use my own yaml to override those sources shipped with magerun with the ones from this PR, so for me it wouldn't hurt if this won't get merged.
Maybe we shouldn't hang too much on those old versions ;-)

ktomk:
@arosenhagen: Thanks! If this works with magedownload-cli I think that would be great. Merged.

PR #767, opened 2016-02-09
Added 'docroot' as a subfolder (when detecting the Magento installation)

erfanimani:
Many installations use `docroot` term to specify the Magento/webapplication installation directory.

I.e.: http://devdocs.magento.com/guides/v2.0/install-gde/basics/basics_docroot.html (I know Magento 2 uses `pub`, but docroot is a popular term - same in Drupal world.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -34,6 +34,7 @@ detect:
     - webroot
     - web-root
     - wwwroot
+    - docroot
 
 event:
   subscriber:


tkn98:
I've never seen that the folder itself is named `docroot`, it is also that with the reference to the devdocs the term *"Docroot"* is explained, but it's not referenced to name the directory as "`docroot`". Can you please provide the reference you give to Drupal that shows the directory is named that way?

cmuench:
@tkn98 I'm not sure if we should add that to dist config? If anyone needs that it's possible to add the directory in a user or system config.

tkn98:
@cmuench: Not sure either, I don't see much reference it's actually a common directory name (only a common name to describe a specific - but differently named - directory on disk).

erfanimani:
Ah I just noticed this page: https://github.com/netz98/n98-magerun/wiki/Config, meaning I can add my own directory name :)

Feel free to close - I did some Googling and you're right, it doesn't seem to be a common project subdirectory. A couple companies use the `docroot` folder name for Drupal builds, but it doesn't seem to be standardised.

ktomk:
Let's stash it then for the moment. If this needs traction in the future, we can re-open. It's easy to solve with a local config as well for specific setups.

Issue #771, opened 2016-02-12
db:dump - sed: RE error: illegal byte sequence - mysqldump: Got errno 32 on write


convenient:
#### Error
When running the `magerun db:dump` command I received the following error.

```
sed: RE error: illegal byte sequence
mysqldump: Got errno 32 on write
```

#### Explanation
I traced it back to the cool definer logic introduced in #130 

It turned out to be related to the `$LANG` variable. My OSX machine had `$LANG` defined like:

```
echo $LANG
en_GB.UTF-8
```

#### Workaround

We got the dump command to work by running it like
```
export LANG=C
magerun db:dump
```

Thanks @dank00 for the workaround.

cmuench:
I confirm this behavior on OSX.
If we prepend `LANG=C LC_CTYPE=C LC_ALL=C` before `sed` the problem is gone away.

dank00:
Hi @cmuench busy at work at the moment but I've put this here #773

ktomk:
This is now part of the development version and will be part of the next stable release. Please update to the unstable version and try again. It should now run w/o the error. Please provide feedback if you can, I'll close for now as I'm confident that it works.

convenient:
Bam! If only all issues could be resolved this quickly / easily.

PR #773, opened 2016-02-15
[FIX] sed: RE error: illegal byte sequence on myqldump

dank00:
Prepend LANG=C LC_CTYPE=C LC_ALL=C to sed command

issue: #771


cmuench:
@dank00 Thanks

PR #777, opened 2016-02-17
Refactor mysqldump command order

convenient:
The intention of this is to make it easier to grok when looking at both generated commands.

With this modification it will generate commands like:
```
mysqldump --single-transaction --quick --routines --no-data -h'127.0.0.1'
mysqldump --single-transaction --quick --routines -h'127.0.0.1 --all-ignores-go-here'
```

When you have hundreds of tables to be ignored, quickly grokking the table being dumped can be annoying with the current implementation as you have to view/modify the very beginning of the first line with the very end of the second line.


ktomk:
Superseeded by #778 (wrong branch)

PR #778, opened 2016-02-17
Refactor mysqldump command order

convenient:
The intention of this is to make it easier to grok when looking at both generated commands.

With this modification it will generate commands like:
```
mysqldump --single-transaction --quick --routines --no-data -h'127.0.0.1'
mysqldump --single-transaction --quick --routines -h'127.0.0.1' --ignored-tables-go-here -ad-infinitum
```

When you have hundreds of tables to be ignored, quickly grokking the table being dumped can be annoying with the current implementation as you have to view/modify the very beginning of one line with the very end of one line.


ktomk:
@convenient: Thanks you! The changes are not part of the development version and will become part of the next stable release.

PR #787, opened 2016-03-10
Fix casting array_combine() result to array

### LINKED ISSUE #786: Hiding underlying error condition in #785
Just a quick note that the change 5670cb0d74cbb03241f98f74d868a788d88fb80a (#785) hides the error condition. Both return points in [`Cron\AbstractCronCommand::getSchedule()`](https://github.com/netz98/n98-magerun/blob/5670cb0d74cbb03241f98f74d868a788d88fb80a/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php#L47) allow to return `FALSE` which is unchecked (cause of original error) however signals a problem with that (yet still unfixed) code.

Casting `FALSE` to array btw creates `[FALSE]` which is not semantically wrong operation in the now "fixed" code.

I'd say this needs proper fixing.

/cc @dmaciej: FYI

dmaciej:
Resolves #786

### DIFF
--- a/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php
+++ b/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php
@@ -32,7 +32,12 @@ abstract class AbstractCronCommand extends AbstractMagentoCommand
 
         foreach ($jobs as $name => $job) {
             /* @var $job Mage_Core_Model_Config_Element */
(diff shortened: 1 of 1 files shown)


ktomk:
@dmaciej I don't know if this is a fix solution yet, I think the error is more underlying in a different component and needs fixing there. Thanks for your efforts though, let's keep it for discussion in the new issue first. I haven't yet a probable fix suggestion so you can blame me on this, too :)

Issue #799, opened 2016-03-25
Magerun plugins not loaded anymore


therouv:
In the new version `1.97.15` all custom magerun plugins are not loaded anymore. 

The bug was introduced with commit https://github.com/netz98/n98-magerun/commit/1f0b0edcdf1e9a2ca3a04230ed83975afca2808e#diff-38c7bfd66e260842fd84690f151b29d8L218 where the plugin folders were changed to **n98-magerun2** which seems to be wrong since this repository is magerun for Magento1.

ktomk:
Yes, this was a mistake while streamlining both repos with each other.

Has been addressed in 1.97.16.

Thanks for the report + patch!

PR #800, opened 2016-03-25
Fix magerun plugin loader

therouv:
Fixes bug which was introduced in latest magerun version. 
See #799 for more information.


ktomk:
@therouv:; Thx for the quick patch! Will file the next release version.

ktomk:
Magerun version 1.97.16 with this fix has been released. Sorry for the interruption.

therouv:
@ktomk Awesome, thanks!

Issue #811, opened 2016-04-12
scope id 


nathan333:
Hi there, 

I just found something that might be a small issue with config.  In src/N98/Magento/Command/Config/AbstractConfigCommand.php there was a change. Im guessing it was line 117     if (0 >= (int)$scopeId) . I noticed this because I have be trying to run config:set --scope websites --scope-id 0 blah blah, and it was working fine last week but not today (i saw a commit to this file 5 days ago). I also noticed on your readme that the default scope id is 0, however when that is run you get that error.  

Just a heads up!

tkn98:
Hi Nathan, thanks for the report. It is correct that an error is given since a more recent version.

I belief that we should not use zero (0) within the *websites* scope. This is why you see the error. This was left unchecked in previous versions and is now more strict.

If you think this is an error, it would be great you could show a valid example or some other kind of reference that shows how a scope-id of 0 is used within the *websites* scope.

rjocoleman:
I've got the same issue. I was previously using `config:delete 'web/secure/base_url' --scope-id 0 --scope='stores'`.

I have the `base_url`'s set for admin store, scope 0.

This is not a GUI accessible configuration but it works, I do not remember the context around setting it to work this way as it was many years ago, it may well be an abuse of the core scope handling.

The new `config:delete` validation specifically disables this use case. I saw the `--all` switch as a mitigation for this but it doesn't delete _all_ config per #814.

nathan333:
yeah. I work with a few customers and they all have their admin as scope 0. Its a little frustrating as I cannot really change this for them. I have just reverted back to using the older version.

rjocoleman:
I see a few options:

1. #814 (`all` command using default too), this is potentially imprecise for this specific issue. Arguably `all` should do _all_ but that's a separate topic for #814)
2. Less validation.
3. Another flag to force validation to be disregarded - non-default, for those that need it - `config:delete --force --scope-id 0 --scope default`.
4. Out of core custom command to handle this use case with less validation e.g. `config:delete:force --scope-id 0 --scope default`.

I'm happy to whip up any of these options and send the appropriate PR (or create a gist for a custom command).

nathan333:
I don't mind its up to you. I will just continue to use the older version. I only need it for the next couple of weeks

ktomk:
To not lock anybody out I then suggest the variant with the `--force` switch to allow 0 for scope stores and websites.

[...]

PR #814, opened 2016-04-13
include default website/store in config:delete

rjocoleman:

~~- [] README.md reflects changes (if any)~~ **N/A**

Subject: include default website/store in config:delete

Per README `config:delete --all` _"Deletes all entries of a path (ignores --scope and --scope-id)"_

As implemented `config:delete --all` enumerates `Mage::app()->getWebsites()` and `Mage::app()->getStores()`.

With no parameters passed the default website and store is excluded from the respective `Mage::app()->getWebsites()` and `Mage::app()->getStores()` functions.

Per https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Core/Model/App.php#L984-L1002 and https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Core/Model/App.php#L891-L914

### DIFF
--- a/src/N98/Magento/Command/Config/DeleteCommand.php
+++ b/src/N98/Magento/Command/Config/DeleteCommand.php
@@ -86,12 +86,12 @@ HELP;
             $deleted[] = $this->deleteConfigEntry($path, 'default', 0);
 
             // Delete websites
-            foreach (\Mage::app()->getWebsites() as $website) {
+            foreach (\Mage::app()->getWebsites(true) as $website) {
                 $deleted[] = $this->deleteConfigEntry($path, 'websites', $website->getId());
             }
 
             // Delete stores
-            foreach (\Mage::app()->getStores() as $store) {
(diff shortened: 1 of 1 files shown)


ktomk:
Thanks for the suggestion, but [in Line 68 the default entry (id 0) is add](https://github.com/rjocoleman/n98-magerun/blob/patch-1/src/N98/Magento/Command/Config/DeleteCommand.php#L86) for the all deletion.

From what I know for config and scopes is that you can't make a config setting for scope *websites* and *stores* with ID zero (0).

If you think that is an error, I would like to see this discussed. The topic was just raised in #811 as well (there it is about verification of input parameters), perhaps you can take a look there, too.

rjocoleman:
thanks @ktomk I'll continue the discussion over on #811 as it's quite similar.

rjocoleman:
closing in favour of #815

PR #815, opened 2016-04-13
Add config:delete --force option

### LINKED ISSUE #811: scope id

rjocoleman:
Subject: Add config:delete --force option

Fixes #811 #814.

As discussed in #811 and #814 the recent changes in 77df9c2b7110b370cfb6608aed333e0777951622 prevent some users who have set non-standard `scope-id` and `scope` combinations from using `config:delete`.

Additionally if these _non-standard_ `scope-id` and `scope` combos are used `--all` isn't true to it's documentation. 

This PR adds a `--force` option to `config:delete` to avoid this lock-out. This option does two things:

1. When used without the `--all` option: the _"positive integer"_ validation of `scope-id` is skipped, allowing usage of `0` `scope-id` for both `websites` and `stores` scopes.

2. When used with the `--all` option: `true` is passed to [`Mage::app()->getStores()`](https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Core/Model/App.php#L891-L914) and [`Mage::app()->getWebsites()`](https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Core/Model/App.php#L984-L1002) to include default stores and websites in the enumeration. This makes `--all` true to it's documentation.

**Update**

`config:set` also uses the same validation logic. https://github.com/netz98/n98-magerun/pull/815/commits/666b5f8b9d6140a2b01c97df011909a6b5d6ed35 adds a `--force` option to `config:set` to as in 1. above.

### DIFF
--- a/src/N98/Magento/Command/Config/AbstractConfigCommand.php
+++ b/src/N98/Magento/Command/Config/AbstractConfigCommand.php
@@ -71,10 +71,11 @@ abstract class AbstractConfigCommand extends AbstractMagentoCommand
     /**
      * @param string $scope
      * @param string $scopeId
+     * @param boolean $allowZeroScope
      *
      * @return string non-negative integer number
      */
-    protected function _convertScopeIdParam($scope, $scopeId)
+    protected function _convertScopeIdParam($scope, $scopeId, $allowZeroScope=false)
     {
         if ($scope === 'default') {
             if ("$scopeId" !== "0") {
@@ -114,10 +115,12 @@ abstract class AbstractConfigCommand extends AbstractMagentoCommand
             );
         }
 
-        if (0 >= (int)$scopeId) {
-            throw new InvalidArgumentException(
-                sprintf("Invalid scope parameter, %s is not a positive integer value", var_export($scopeId, true))
-            );
+        if (!$allowZeroScope) {
+            if (0 >= (int)$scopeId) {
+                throw new InvalidArgumentException(
+                    sprintf("Invalid scope parameter, %s is not a positive integer value", var_export($scopeId, true))
+                );
+            }
         }
 
         return $scopeId;
--- a/src/N98/Magento/Command/Config/AbstractConfigCommand.php
+++ b/src/N98/Magento/Command/Config/AbstractConfigCommand.php
@@ -115,12 +115,10 @@ abstract class AbstractConfigCommand extends AbstractMagentoCommand
             );
         }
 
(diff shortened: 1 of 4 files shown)


rjocoleman:
I've left this a two commits for now since it's doing two things. Happy to rebase, rework the solution etc before merge.

src/N98/Magento/Command/Config/AbstractConfigCommand.php
```diff
@@ -114,10 +115,12 @@ protected function _convertScopeIdParam($scope, $scopeId)
             );
         }
 
-        if (0 >= (int)$scopeId) {
-            throw new InvalidArgumentException(
-                sprintf("Invalid scope parameter, %s is not a positive integer value", var_export($scopeId, true))
-            );
+        if (!$allowZeroScope) {
+            if (0 >= (int)$scopeId) {
```
ktomk: `0 - (bool)$allowZeroScope >= (int)$scopeId`

this will allow the zero scope then. but not -1 (etc). it's also one less if.

ktomk:
Two commits are fine. I've left a small comment on one of the two commits. Otherwise this is fine to merge.

rjocoleman:
Thanks @ktomk I've revised that

rjocoleman:
@ktomk any desire for the same `--force` option on `config:set`?

I don't use it personally but I can see how others could be relying on this workflow that's now blocked.

The same abstract function is used so it's a very simple addition.

ktomk:
@rjocoleman: Good idea, looks missed so far. Please add it while it's hot.

rjocoleman:
done! https://github.com/netz98/n98-magerun/pull/815/commits/666b5f8b9d6140a2b01c97df011909a6b5d6ed35

ktomk:
@rjocoleman: Thanks! Merged and shipped with 1.97.19.

rjocoleman:
thanks @ktomk it's working great for me 👍

PR #816, opened 2016-04-16
Fixes a rare PHP notice in the dev:module:list command.

hostep:
### Magerun pull-request check-list:

### Subject:

Fixes a rare PHP notice in the `dev:module:list` command.

### Changes proposed in this pull request:

One of the modules in our shops has this module xml file:
```xml
<config>
    <modules>
        <!-- disable old Ogone extension -->
        <Mage_Ogone>
            <active>false</active>
        </Mage_Ogone>

        <Netresearch_OPS>
            <active>true</active>
            <codePool>community</codePool>
            <depends>
                <Mage_Sales/>
                <Mage_Payment/>
            </depends>
        </Netresearch_OPS>
    </modules>
</config>
```

The `Mage_Ogone` module referenced here was initially included in older Magento versions (like CE 1.4 and CE 1.5 if I remember correctly).
The `Netresearch_OPS` module integrates with the same Payment Provider, and they decided to disable the default module which came with Magento.
In newer versions of Magento, the `Mage_Ogone` is no longer is included.

So, on a newer Magento version, if you run the `dev:module:list` command, it will find the `Mage_Ogone` module, but since it hasn't got a `codePool` defined, we get this PHP notice in the `system.log` file:
`ERR (3): Notice: Undefined index: codePool  in n98-magerun.phar/src/N98/Magento/Command/Developer/Module/ListCommand.php on line 67`

This PR fixes this notice, by first checking if the module definition has at least a `codePool` and `active` flag defined, otherwise it will skip the module in the output. I think this makes sense, because if neither of those definitions are set, the module won't be active in Magento and it doesn't make sense to output it in the module definition.
If you disagree with this, let me know, maybe we can still output the module, but with a warning about it missing some definitions or something like that.

I know this is a very rare case, so if you want to ignore this PR, no problem!

### DIFF
--- a/src/N98/Magento/Command/Developer/Module/ListCommand.php
+++ b/src/N98/Magento/Command/Developer/Module/ListCommand.php
@@ -63,21 +63,15 @@ class ListCommand extends AbstractMagentoCommand
     {
         $modules = \Mage::app()->getConfig()->getNode('modules')->asArray();
         foreach ($modules as $moduleName => $moduleInfo) {
-
-            // skip modules which don't have a codePool or an active definition
-            if (!array_key_exists('codePool', $moduleInfo)
-                || !array_key_exists('active', $moduleInfo)
-            ) {
-                continue;
-            }
+            $codePool = isset($moduleInfo['codePool']) ? $moduleInfo['codePool'] : '';
+            $version  = isset($moduleInfo['version'])  ? $moduleInfo['version']  : '';
+            $active   = isset($moduleInfo['active'])   ? $moduleInfo['active']   : '';
 
             $this->infos[] = array(
-                'codePool' => $this->sanitizeModuleProperty($moduleInfo['codePool']),
+                'codePool' => $this->sanitizeModuleProperty($codePool),
                 'Name'     => $this->sanitizeModuleProperty($moduleName),
-                'Version'  => isset($moduleInfo['version'])
-                    ? $this->sanitizeModuleProperty($moduleInfo['version'])
-                    : '',
-                'Status'   => $this->formatActive($moduleInfo['active']),
+                'Version'  => $this->sanitizeModuleProperty($version),
+                'Status'   => $this->formatActive($active),
             );
         }
     }
--- a/src/N98/Magento/Command/Developer/Module/ListCommand.php
+++ b/src/N98/Magento/Command/Developer/Module/ListCommand.php
@@ -63,6 +63,14 @@ class ListCommand extends AbstractMagentoCommand
     {
         $modules = \Mage::app()->getConfig()->getNode('modules')->asArray();
         foreach ($modules as $moduleName => $moduleInfo) {
+
+            // skip modules which don't have a codePool or an active definition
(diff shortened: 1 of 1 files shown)


ktomk:
Dropping the module from the list seems a bit harsh to me and I think that zero-length strings are fine if there are no values. 

The warning is indeed a flaw of that part in the code as array entries are used without checking first if expected entries do exist.

src/N98/Magento/Command/Developer/Module/ListCommand.php
```diff
@@ -63,6 +63,14 @@ protected function findInstalledModules()
     {
         $modules = \Mage::app()->getConfig()->getNode('modules')->asArray();
         foreach ($modules as $moduleName => $moduleInfo) {
+
+            // skip modules which don't have a codePool or an active definition
+            if (!array_key_exists('codePool', $moduleInfo)
```
ktomk: `!array ...` should be on a new line.

hostep:
Well the question is: what will cause most confusion for developers?

1. If the module is displayed in the list, but actually there is no module in the first place (in my example that's the case)
2. If the module is not displayed in the list, because a developer is working on a module and has accidentally forgotten to define the `codePool` or `active` parameter.

I can see valid reasons for the two cases, maybe the first one is best, since the `dev:module:list` command is meant for development debugging.

Is it ok if I update my PR and do the same check as is now the case with the `version` parameter?

Then this would be the result, in the example above:
```
+-----------+------------+------------+----------+
| codePool  | Name       | Version    | Status   |
+-----------+------------+------------+----------+
|           | Mage_Ogone |            | inactive |
+-----------+------------+------------+----------+
```

hostep:
There you go, I already updated the PR :)

I also wanted to change the capitalization of `codePool` to `CodePool`, to match the other headers, but I don't know if this is considered breaking backwards compatibility?

ktomk:
Good catch! I think Code-Pool is correct English for headers. Correct? Do as you see fit. Otherwise patch looks good to merge!

hostep:
There you go. I decided to go for `Code pool` since I found two other places where it is being used like that:
* https://github.com/netz98/n98-magerun/blob/36e8c5c14a021a473a1b5efe43bbf2fc132f7d87/src/N98/Magento/Command/Developer/Module/Dependencies/AbstractCommand.php#L73
* https://github.com/netz98/n98-magerun/blob/36e8c5c14a021a473a1b5efe43bbf2fc132f7d87/src/N98/Magento/Command/Developer/Module/Dependencies/OnCommand.php#L63

ktomk:
@hostep: Thanks! And +1 for checking for consistent wording across magerun!

ktomk:
Changing the wording in 1.97.20 did break athrid party module. I'll revert the wording in 1.97.21. We need to keep the change for future.

hostep:
Ok, I understand the reasoning (I already was afraid it would break backwards compatibility), best to wait to make the `Code pool` header consistent across the project in version 1.98 then, no problem :)

PR #817, opened 2016-04-17
[#23] Add notice of developer IP restrictions when toggling inline tr…

### LINKED ISSUE #23: Check if developer IP restriction is set / allow reset of the option
When enabling template hints or inline translation, the setting might not take effect because a developer IP restriction is set

* show warning
* add an option to reset the IP

robbieaverill:

Fixes #23 

- Add a set of commands to `AbstractMagentoStoreConfigCommand`
- Detect if developer IP restrictions are enabled when changing template hints or inline translations
- Present the current value and ask if it should be changed
- Change if required

Questions:

- Is this the best place for these commands to be stored? I'm not sure that it is. Perhaps a new `DeveloperHelper` or something?

### DIFF
--- a/src/N98/Magento/Command/AbstractMagentoStoreConfigCommand.php
+++ b/src/N98/Magento/Command/AbstractMagentoStoreConfigCommand.php
@@ -95,7 +95,7 @@ abstract class AbstractMagentoStoreConfigCommand extends AbstractMagentoCommand
      * @param InputInterface $input
      * @param OutputInterface $output
      *
-*@return int|void
+     * @return int|void
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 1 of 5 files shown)


robbieaverill:
I'll be surprised if this doesn't break some of the unit tests for the commands I changed, and will either fix or add new tests for this command tomorrow morning. In the meantime Tom if you have time could you give some feedback on the method placement? Cheers :)

ktomk:
@robbieaverill: It doesn't break any test ;) - Perhaps not yet tested? ^^ - For the placement: I think for the first implementation it's okay, albeit I would favour as well composition over inheritance. But this could be extracted later on. Have you tried what happens in non-interactive mode? I just did and **\N98\Util\Console\Helper\ParameterHelper** has a problem when asking for store in that mode. That would btw. be an example of an existing helper but I don't like this much. Perhaps better just a model for that job (standard PHP class next to the command) but as written, I don't think it's needed for the first implementation.

ktomk:
@robbieaverill: Can you squash the two commits into one and shorten the subject-line of that commit to 50 characters?

This is the only two things that prevented me to merge your PR right away. Thanks!

robbieaverill:
@ktomk done. I'm happy to add unit tests, but can do it in a separate PR if you'd like.

ktomk:
@robbieaverill: Thanks! Tests are always welcome, just create a new PR if you like.

PR #818, opened 2016-04-21
Adds command to list transactional email template usage.

MikeParkin:
Adds command to list transactional email template usage.

When an installation has many transactional emails, it is easy
to lose track of which ones are currently being used. As you can
also fall back to using templates in the source code, it is helpful
to be able to see which email templates are in use and which can be deleted.

This command allows you to see which email templates are used, and in
which scopes of the configuration that is.

I've not added unit tests, let me know if these are required and I'll get them written.

### DIFF
--- a/src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
+++ b/src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace N98\Magento\Command\Developer\EmailTemplate;
+
+use N98\Magento\Command\AbstractMagentoCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use N98\Util\Console\Helper\Table\Renderer\RendererFactory;
+use Mage;
+use Mage_Adminhtml_Model_Email_Template;
+
+class UsageCommand extends AbstractMagentoCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('dev:email-template:usage')
+            ->setDescription('Display database transactional email template usage')
+            ->addOption(
+                'format',
+                null,
+                InputOption::VALUE_OPTIONAL,
+                'Output Format. One of [' . implode(',', RendererFactory::getFormats()) . ']'
+            )
+        ;
+    }
+
+    /**
+     * @param InputInterface  $input
+     * @param OutputInterface $output
+     *
+     * @return int|void
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->detectMagento($output, true);
+        $this->initMagento();
+        $templates = $this->findEmailTemplates();
+
+        if (!empty($templates)) {
+            $this->getHelper('table')
+                ->setHeaders(array('id', 'Name', 'Scope', 'Scope Id', 'Path'))
+                ->renderByFormat($output, $templates, $input->getOption('format'));
+        } else {
+            $output->writeln("No transactional email templates stored in the database.");
+        }
+    }
+
+    protected function findEmailTemplates()
+    {
+        $templates = Mage::getModel('adminhtml/email_template')->getCollection();
+
+        $return = array();
+
+        foreach ($templates as $template) {
+
+            /**
+             * Some modules overload the template class so that the method getSystemConfigPathsWhereUsedCurrently
+             * is not available, this is a workaround for that
+             */
+            if (!method_exists($template, 'getSystemConfigPathsWhereUsedCurrently')) {
+                $instance = new Mage_Adminhtml_Model_Email_Template();
+                $template = $instance->load($template->getId());
+            }
+
+            $configPaths = $template->getSystemConfigPathsWhereUsedCurrently();
+
+            if (!count($configPaths)) {
+                $configPaths[]  = array(
+                    'scope'         => 'Unused',
+                    'scope_id'      => 'Unused',
+                    'path'          => 'Unused',
+                );
+            }
+
+            foreach ($configPaths as $configPath) {
+                $return[] = array(
+                    'id'            => $this->sanitizeEmailProperty($template->getId()),
+                    'Template Code' => $this->sanitizeEmailProperty($template->getTemplateCode()),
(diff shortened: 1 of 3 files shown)


src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
```diff
@@ -0,0 +1,101 @@
...
+             * is not available, this is a workaround for that
+             */
+            if (!method_exists($template, 'getSystemConfigPathsWhereUsedCurrently')){
+                $instance = new \Mage_Adminhtml_Model_Email_Template();
+                $template = $instance->load($template->getId());
+            }
+
+            $configPaths = $template->getSystemConfigPathsWhereUsedCurrently();
+
+            if (count($configPaths)){
```
ktomk: Alternative: If there is no configPaths entry you could create a single entry with the three "unused" strings for scope, scope_id and path.

then you could *always* foreach.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $templates = \Mage::getModel('adminhtml/email_template')->getCollection();
+
+        foreach ($templates as $template){
+
+            /**
+             * Some modules overload the template class so that the method getSystemConfigPathsWhereUsedCurrently
+             * is not available, this is a workaround for that
+             */
+            if (!method_exists($template, 'getSystemConfigPathsWhereUsedCurrently')){
```
ktomk: Nice internal. Please import the magento core class (via use statment) so that it does not need a trailing \ here.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
```diff
@@ -0,0 +1,101 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use N98\Util\Console\Helper\Table\Renderer\RendererFactory;
+
+class UsageCommand extends AbstractMagentoCommand
+{
+    /**
+     * @var array
+     */
+    protected $infos;
```
ktomk: Why protected? And have you considered to just return this as an array by the function that builds it?

ktomk:
Thanks a lot for your PR. I like this command suggestion. I have left some comments with your suggested commit, it's more informative so if you see something in the feedback, please change as you see fit or just provide feedback yourself to my comments.

About the tests you ask about: Tests are always welcome, you can try if you do a happy path check for that command with the command tester, but I do not think it's an obligation. It's perhaps worth for your own to see which paths are covered by a simple command test. Let me know if you're not familiar with the command tester so that I can link you an example usage in n98-mageurn.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
MikeParkin: Yep good idea, I'll reformat as suggested - will make it easier to read.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
MikeParkin: Yeah, cool function huh. Happy to import with 'use', is this the preferred method? I followed examples in other classes and there are currently many uses of \Mage::

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
MikeParkin: Used the dev:module:list as a starting point, so copied from there - you're right though, no real need to cache this in a protected member - will update.

MikeParkin:
Great, glad you like it. I've made the changes as discussed. 

Thanks for information regarding unit tests, hopefully take a look over the weekend.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
ktomk: Yes, correct observation. We don't have this consistently, but whenever I touch the code I change it to *make use of use*. That helps to keep track which other classes are in use within a certain file.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
ktomk: Thanks for the hint to dev:module:list, I'll do the correction there then after merging your PR.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
```diff
@@ -61,32 +60,34 @@ protected function findEmailTemplates()
...
-                        'Template Code' => $this->sanitizeEmailProperty($template->getTemplateCode()),
-                        'Scope'         => $this->sanitizeEmailProperty($configPath['scope']),
-                        'Scope Id'      => $this->sanitizeEmailProperty($configPath['scope_id']),
-                        'Path'          => $this->sanitizeEmailProperty($configPath['path']),
-                    );
-                }
-            } else {
-                $this->infos[] = array(
+            if (!count($configPaths)){
+                $return[]  = array(
```
ktomk: Just add to $configPaths the first entry (instead of to $return). You would only need to create three entries then (the three "unsused" ones). The rest then will just work as follows.

ktomk:
@MikeParkin: Please squash the changes into a single commit and force-push. That will update the PR.

src/N98/Magento/Command/Developer/EmailTemplate/UsageCommand.php
MikeParkin: Sorry not following on this one, could you explain what you mean?

ktomk: Sure, can you see this: https://github.com/MikeParkin/n98-magerun/compare/develop...ktomk:patch-1 ? Instead of manipulating $return, it manipulates the $configPaths (for the unused case).

MikeParkin: Ah yes, now this makes perfect sense!

MikeParkin:
Done!

ktomk:
@MikeParkin: Thanks!

PR #825, opened 2016-05-17
Additional tables in customer data strip

moloughlin:
Subject: Add two tables to be stripped when using the @customers directive 

Fixes: 
Fatal error (FK constraint violations) when using the Enterprise backup functionality on a database which was dumped by N98 with customer data stripping. The salesrule_coupon_usage and salesrule_customer tables both reference customer_entity.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -512,7 +512,7 @@ commands:
 
       - id: customers
         description: Customer data - Should not be used without @sales
-        tables: customer_address* customer_entity* wishlist*
+        tables: customer_address* customer_entity* wishlist* salesrule_coupon_usage salesrule_customer
 
       - id: newsletter
         description: Newsletter subscriber data
--- a/config.yaml
+++ b/config.yaml
@@ -508,11 +508,11 @@ commands:
 
       - id: sales
         description: Sales data (orders, invoices, creditmemos etc)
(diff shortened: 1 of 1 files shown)


tkn98:
@moloughlin: Thanks for the PR. From a quick look, this perhaps belongs into *`@sales`* (see as well the description of *`@customer`*)?

moloughlin:
Thanks @tkn98, fair point. I've moved them into the sales set since typical usage of `@customer` would include `@sales` anyway.

tkn98:
@moloughlin: Did you run a test-export with CE edition as well so that it's clear it doesn't break anything. Sorry if this sounds lazy from my end, but I'm not sure what happens if the table does not exists and I can't review right now.

moloughlin:
@tkn98 Tested a dump/re-import of clean CE. Please note that the tables added are still part of the CE, but this particular error isn't actually hit unless using some EE-specific functionality.

ktomk:
@moloughlin; Thanks a lot, I've polished the commit a bit, it's now part of the develop branch. I was just seeing the merge commit was coming from two versions ago. Just asking, is there any reason you based on that version, e.g. is there something broken for you with the current ones or was there no particular reason?

ktomk:
@moloughlin: Released with 1.97.22

PR #837, opened 2016-07-04
FIX strip email queue tables for development db:dump commands

### LINKED ISSUE #836: db:dump: @development should strip the email queue
Presently it doesn't strip the `core_email_queue` tables, it should so that we don't end up with real customer email data in development environments.

robbieaverill:
Fixes #836 

- Add `@emails` db:dump group
- Add it to `@development`

* Resolves #836


ktomk:
Thanks!

PR #838, opened 2016-07-11
Adding model::method to cron list command

steverobbins:

- Adds `Model` column to `sys:cron:list` command, showing the model::method that the job uses.

e.g.

before

changes to

after


src/N98/Magento/Command/System/Cron/AbstractCronCommand.php
```diff
@@ -22,7 +22,11 @@ protected function getJobs()
         $jobs = $this->getJobConfigElements();
 
         foreach ($jobs as $name => $job) {
-            $table[$name] = array('Job' => $name) + $this->getSchedule($job);
+            $model = null;
+            if (isset($job->run) && isset($job->run->model)) {
```
ktomk: A single isset is doing the job here, see for example: https://3v4l.org/nA8Sl

ktomk:
Thanks! For that one comment, I fix that in develop, just FYI.

PR #848, opened 2016-08-03
Add include parameter to db:dump command.

jarodhayes:
- Add an `--include` parameter to the `db:dump` command to dump only specific tables, which operates much like the existing `--exclude` parameter.


ktomk:
Hi Jarod, thanks for your PR. It's no obligation but please let me know if you want to consider looking into improving the code-coverage.

src/N98/Magento/Command/Database/DumpCommand.php
```diff
@@ -243,6 +244,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
+        if ($input->getOption('exclude') && $input->getOption('include')) {
```
ktomk: Why can't I specify include and exclude together? Should allow me to first exclude a bunch but then add two or three other tables as well.

tkn98:
@jarodhayes: Thanks for providing a pull-request. I asked some days ago:

> Why can't I specify include and exclude together? Should allow me to first exclude a bunch but then add two or three other tables as well.

Can you just provide some feedback so that it's better to understand? The rest looks OK, code coverage could be improved but I don't know how much you're into that. Your feedback is welcome!

jarodhayes:
@tkn98 I apologize for the slow response. I think both parameters could be included. I made them mutually exclusive due to ease of implementation more than anything else, as I was unsure how having both parameters should work (e.g. which should take precedence, should parameter order matter, etc). In my personal case, I needed a way to dump and then import my `core_config_data` before and after a database import in my development environment. I thought the feature might be something others find of use, and made sense to me to be included in the core functionality.

Regarding testing, I would be interested in improving code coverage. I am not very familiar with PHPUnit, but I am interested in learning.

tkn98:
Ease of implementation is fine. For tests I have to take a look and it depends a bit on my time. On first glance I see two tests:

1. Test with the new parameter (happy path)

2. Test with the new parameter providing the other one that the exception is thrown

A testcase for the command should exist already, you would add new test methods there.

You start phpunit via `vendor/bin/phpunit`. This might require a local setup. For reference please also see this recent comment: https://github.com/netz98/n98-magerun/issues/878#issuecomment-266542078 which contains a listing of the various ways on how to get your source code to run incl. local setup, tests and build.

Please provide feedback if the information so far is helpful for you.

jarodhayes:
Thank you, that was very helpful for getting PHPUnit setup. I added some tests that hopefully cover the usage of the `--include` and `--exclude` parameters.

ktomk:
Thank you for contributing to n98-magerun! The new --include feature is now part of the unstable version, you can run the `self-update` command with the `--unstable` parameter or in your case checkout the development branch and pull in latest changes.

PR #849, opened 2016-08-08
Add command eav:attribute:create-dummy-values (with tests)

giuseppemorelli:
Hi guys,
here I am with a new PR from my addons.
Hope my tests are good and the code quality is better than before.

- add a command **eav:attribute:create-dummy-values** to create dummy values for dropdown attributes
- used Faker library to create dummy values

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -1424,6 +1424,29 @@ Create dummy categories with all default vanilla magento or your custom values.
 | `category-name-prefix`       | Category Name Prefix (default: 'My Awesome Category')                                       | any                                              |
 +------------------------------+---------------------------------------------------------------------------------------------+--------------------------------------------------+
 
+Create dummy Dropdown Attribute Values
+""""""""""""""""""""""""""""""""""""""
+
+.. code-block:: sh
+
+   $ n98-magerun.phar eav:attribute:create-dummy-values
+
+Create dummy attribute values (ONLY FOR DROPDOWN ATTRIBUTE)
+
+**Interactive mode** or via **shell arguments** or mixed.
+
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+| Arguments                    | Description                                  | Accepted Values                                              |
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+| `locale`                     | Locale value in ISO standard like en_US      | only string                                                  |
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+| `attribute-id`               | Attribute ID to add values                   | only integer                                                 |
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+| `values-type`                | Types of Values to create (default int)      | `int`<br />`string`<br />`color`<br />`size`<br />`designer` |
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+| `values-number`              | Number of Values to create (default 1)       | only integer                                                 |
++------------------------------+----------------------------------------------+--------------------------------------------------------------+
+
 
 List Extensions
 """""""""""""""
--- a/src/N98/Magento/Command/Eav/Attribute/Create/DummyCommand.php
+++ b/src/N98/Magento/Command/Eav/Attribute/Create/DummyCommand.php
@@ -1,0 +1,374 @@
+<?php
+
+namespace N98\Magento\Command\Eav\Attribute\Create;
+
+use N98\Magento\Application;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
+
+class DummyCommand extends \N98\Magento\Command\AbstractMagentoCommand
+{
+    protected $sizes = array(
+        'XXS', 'XS', 'S', 'M', 'L', 'XL', 'XXL', 'XXXL', '37', '38',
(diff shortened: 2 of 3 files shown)


ktomk:
@giuseppemorelli: Thanks a lot! If you're interested in the code-quality, you can find some inspections on scrutinizer-ci. Don't be shocked on the sheer number of issues your PR introduced, if you look inside you'll understand why (and that it's merely two, three issues and not 113 as the number shows: https://scrutinizer-ci.com/g/netz98/n98-magerun/inspections/ef017c6c-cdd9-429f-9aac-7bd0e5b955ad)

ktomk:
@giuseppemorelli: I just reviewed your code. There are only minor issues, I fixed the ones I saw (you'll find a link to the diff at the end of my comment) here is the more likely more informative stuff listed, pick what is interesting for you, question what sounds contradicting to you and about what you want to learn more:

* `\r\n` is not needed for `output->writeln` as it will write a line already and therefore does the break at the end (to denote a line-break). Additionally `\r\n` is not cross-platform for line-breaks, writeln will take care of that, too. We just do not need to add it.

* In a text-file each line is terminated with the line-termination character ("`\n`"). This is standard, and we use it for text-files which PHP files are as well. This is true for the last line of the file you have in your commit as well. Yours did miss the line-terminator on the last line. So take care to terminate every text-file with a new-line ("`\n`").

* You should not make use of \` (backtick) to do single-quotes within PHP code as it is [the execution operator](http://php.net/manual/en/language.operators.execution.php). The "you should not" is meant indirect, this can have security implications therefore should be handled with respect and distance. It's much easier and cleaner to just use single quotes ("`'`") instead.

* There is no need to pass objects by reference as object parameters. Albeit it even does not directly hurt, there is a problem with this practice: Passing parameters by reference signal return values and/or side-effects. That's not your case here. As passing parameters by reference are an expensive operation per-se, and you have no need for this practice, just do not pass per reference. Instead whenever you feel the urge to pass a parameter by reference, try instead to not pass it by reference at all and see if it works. If so, spare the reference. Additionally, always add the type-hint in front (e.g. see manageArguments()).

* Always define properties or methods as private unless there is reason to not do that (e.g. see execute()). You seem to prefer *protected*, but that preference signals unfinished work albeit your commit is working already. I changed all protected ones to private - with only a single exception. It works fine. Therefore this is proof that protected was not necessary and of no use. Take care that *protected* does not enable you for future changes but instead it blocks them. If you see protected, read it as public. Did you really meant to make this all public? No you didn't, I know ;).

* Whenever you use broad terms like "manage" in method or variable names, see it as a red flag. You most likely can name better (more concrete) what this method or variable is about. Do so, it will help you reading and understanding your and other persons code over a longer period of time (not fixed, see manageArguments()).

* Writing ALL-CAPS is normally considered screaming. I'm sure every of your comments is important as every of your lines of code is. Comments should not be more important than the code. There really is no need to "scream" in comments.

So much for the review, I bet if you look closely you'll spot the one or other thing your own but nevertheless I hope it was interesting for you. Please ask any kind of questions if you like, especially if the correctness of what I commented is not clear to you. I'm happy to explain or add more background info on any point you'd like to learn more about.

The resulting diff of this comments review is:  729d2414b1b769eba7b9475de0b10468718f1a02 - Please review it and if you spot any kind of error just let me know.

giuseppemorelli:
@ktomk thanks a lot for your advices! I will learn something new with share code with others :)
About your changes:
1) ok, roger that
2) I think my "reformat code" action on phpstorm has deleted it. I'll check and keep in my mind.
3) ok got it
4) at the beginning in my test the command didn't work, so I added reference. Maybe I have missed something or don't know what...
5) ok good point
6) I'll make it better :)
7) Usually I write uppercase comments for separating "blocks" or "sections". How do you write comment to make a comment like a "separator"?

Also, if you don't mind, I have some questions:
A1) Do you add Magento as vendor with namespace Mage?
A2) Why you add class type `InputInterface` in the function arguments? It is in `@param` comment. Is it better in this way for something?

```
/**
     * Manage console arguments
     *
     * @param InputInterface $input
     * @param OutputInterface $output
     *
     * @return array
     */
    private function manageArguments(InputInterface $input, OutputInterface $output)
```
A3) Can I send you new a review with all these advices for `eav:attribute:create-dummy-values` and the old one `category:create:dummy`? So you have new "good" classes.

Thanks a lot for your support!

tkn98:
4) Maybe, I can't say :) All I know is that references should not be necessary for passing objects as parameters in PHP.
7) I don't use comments for separators. If you need to separate lines, use an empty line. That works pretty well.

A1) No I don't do. You have something in mind about that?

A2) @param is informative only, the actual type-hint ensures that you can only pass that one. Using paramter typehints makes code more stable (the one in the method head, not in the docblock comment ;)).
A3) Well, while I was going through the file I already made changes. You can see them in the diff. You can create a new PR based on the current development version if you like. We can also use the Github comments on the actual lines in the code if you like to discuss things further.

Thanks a lot for your support and contribution!

ktomk:
@giuseppemorelli: The code you have submitted contained a die statement. I have not seen this earlier, but now. For the moment I removed the command again, as we can't use "die" or "exit" in the code. This did destroy the testsuite for example.

Commit of the removal is  63237fa.

If you can take a look, I should be able to review it again then in ca. two weeks. Just FYI.

giuseppemorelli:
@ktomk: sorry about that.
I checked in the other commands for best pratice about "catch exception" and didn't find any specific procedure. Just an output warning.
If I remove the "die", the command will be ok or do you find some other errors that make the entire command wrong?

Can I send a new PR with the fix?
Do you have some advice?
Thanks!

ktomk:
@giuseppemorelli: I could not get it to run by just removing the die statement. The build fails on many Magento versions then.

You can always create a new PR. If you do so, it will trigger the build and you can then review it.

On my local test-system everything was fine first.

giuseppemorelli:
@ktomk ok, I'm working on it. I will send you a new PR.
Thanks

tkn98:
Follow-Up is #859

PR #853, opened 2016-08-17
allow environment variables in local.xml

### LINKED ISSUE #851: FR: allow environment variables in local.xml
We currently move our deployment workflow to [envoyer.io](https://envoyer.io). For this to work we developed [Ffuenf_ConfigDotEnv](https://github.com/ffuenf/Ffuenf_ConfigDotEnv) to allow environment variables in local.xml which are loaded via [php-dotenv](https://github.com/vlucas/phpdotenv).
With this in place the n98-magerun db:* commands no longer work since the substitution takes place inside of magento.
To get this going it'll be nice to extend [DbSettings](https://github.com/netz98/n98-magerun/blob/master/src/N98/Magento/DbSettings.php#L88) to allow environment variables in local.xml database settings (see [the implementation](https://github.com/ffuenf/Ffuenf_ConfigDotEnv/blob/master/app/code/community/Ffuenf/ConfigDotEnv/Model/Config/Element.php#L29) for directions).

arosenhagen:
Fixes #851 

use environment variables in local.xml if $value has $-prefix

### DIFF
--- a/src/N98/Magento/DbSettings.php
+++ b/src/N98/Magento/DbSettings.php
@@ -116,6 +116,11 @@ class DbSettings implements ArrayAccess, IteratorAggregate
             $config['unix_socket'] = null;
         }
 
+        // use environment variables if $value has $ prefix
+        foreach ($config as $key => $value) {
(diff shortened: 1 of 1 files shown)


tkn98:
Close as per #851.

PR #862, opened 2016-09-23
Print out cron exceptions

convenient:

I was running a cron via the CLI yesterday and didn't realise it was throwing an exception. This change makes it so that exceptions are output to the command line.

Output Before:

```
Run Namespace_Module_Model_Cron::run
done
```

Output After:

```
Run Namespace_Module_Model_Cron::run
                                                 
  [Mage_Core_Exception]                          
  Exception Message Here
                                                 
sys:cron:run [job]
```

Running the command with `-v` will also print out the stack trace.

### DIFF
--- a/src/N98/Magento/Command/System/Cron/RunCommand.php
+++ b/src/N98/Magento/Command/System/Cron/RunCommand.php
@@ -99,6 +99,7 @@ HELP;
                     ->setMessages($e->getMessage())
                     ->setFinishedAt(strftime('%Y-%m-%d %H:%M:%S', time()))
                     ->save();
+                throw $e;
             }
 
             $output->writeln('<info>done</info>');


tkn98:
Sounds legit. Can you share the exit status code as well? I think it's useful that it is non-zero.

src/N98/Magento/Command/System/Cron/RunCommand.php
```diff
@@ -94,6 +94,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
                     ->setFinishedAt(strftime('%Y-%m-%d %H:%M:%S', time()))
                     ->save();
             } catch (Exception $e) {
+                $output->writeln('');
```
tkn98: There is the \Symfony\Component\Console\Application::renderException method, see src/N98/Magento/Application.php:602 for an example, I think it only needs little adoption for a command.

convenient: Neato, will give it a gander shortly 👍

tkn98:
@convenient: I don't want you to push you into any direction nor burn your time. Just let me know if you want to/can progress with this issue or not. Would be good to know for release planning.

convenient:
Damn, I completely forgot about this with SUPEE8788. Just force pushed up my new change. 

Super small change which works well with the rest of the application, returns a non-zero exit code thanks to the error handling in `Symfony\Component\Console\Application::run`.

tkn98:
Now that's a commit I can't refuse to immediately take in ;) Thanks a lot.

tkn98:
Just re-classified this as a bug as this should provide the non-zero exit code.

convenient:
Thanks dude 👍

PR #863, opened 2016-10-14
Add option to reinitialise cache when cleaning or flushing

arollason:

Subject: Add option to reinitialise cache when cleaning or flushing

Previously the config cache would always be reinitialised when cleaning
or flushing. This was done to resolve a corner case issue with the
changing of database credentials. As the majority of the time no
reinitialisation is required, we move it into a --reinit option, making
cache cleaning much faster, and less prone to error.
    
It should be noted that the method used for reinitialisation of the
config cache will reliably corrupt the config cache when using redis.
There is a "reinit" method in the Mage_Core_Model_Config class which
will reinitialise the config cache and not cause any corruption.However,
instead of changing the method used to rebuild the config cache, this
patch simply puts the automatic rebuild behind an optional flag, as the
config cache reinitialisation is not required when clearing other caches.

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -685,13 +685,17 @@ If you would like to clean only one cache type:
 
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:clean [code]
+   $ n98-magerun.phar cache:clean [--reinit] [--no-reinit] [<code>]
 
 If you would like to clean multiple cache types at once:
 
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:clean [code] [code] ...
+   $ n98-magerun.phar cache:clean [--reinit] [--no-reinit] [<code>] [<code>] ...
+
+Options:
+    --reinit Reinitialise the config cache after cleaning (Default)
+    --no-reinit Don't reinitialise the config cache after cleaning. This will override --reinit.
 
 If you would like to remove all cache entries use `cache:flush`
 
@@ -700,9 +704,15 @@ Run `cache:list` command to see all codes.
 Remove all cache entries
 """"""""""""""""""""""""
 
+Flush the entire cache.
+
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:flush
+   $ n98-magerun.phar cache:flush [--reinit] [--no-reinit]
+
+Options:
+    --reinit Reinitialise the config cache after flushing (Default)
+    --no-reinit Don't reinitialise the config cache after flushing. This will override --reinit.
 
 List Magento caches
 """""""""""""""""""
--- a/src/N98/Magento/Command/Cache/CleanCommand.php
+++ b/src/N98/Magento/Command/Cache/CleanCommand.php
@@ -4,6 +4,7 @@ namespace N98\Magento\Command\Cache;
 
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class CleanCommand extends AbstractCacheCommand
@@ -13,6 +14,18 @@ class CleanCommand extends AbstractCacheCommand
         $this
             ->setName('cache:clean')
             ->addArgument('type', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Cache type code like "config"')
+            ->addOption(
+                'reinit',
+                null,
+                InputOption::VALUE_NONE,
+                'Reinitialise the config cache after cleaning'
+            )
+            ->addOption(
+                'no-reinit',
+                null,
+                InputOption::VALUE_NONE,
+                "Don't reinitialise the config cache after flushing"
+            )
             ->setDescription('Clean magento cache')
         ;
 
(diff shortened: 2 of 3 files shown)


tkn98:
Dear @arollason, thanks a lot for your pull request.

Genereally about the feature you offer here: I can understand you're not confident with the defaults. However we can not change the previous behavior to be backwards compatible.

What I can offer is that we have two flags `--reinit` and `--no-reinit`. The first switch is implicit and technically not needed however allows to override the second one if given so that we can think about a future (if we don't have it yet) local config override for magerun command switch which will give you the outcome you're aiming for while we're maintaining backwards compatibility.

So please preserve the current default behavior and add another switch so that both ways are possible.

arollason:
Hey @tkn98, Thanks for the quick response. It all sounds reasonable. I'll amend the PR on Monday with the suggested changes. Cheers.

tkn98:
@arollason: Thanks a lot, just take your time. If you need any feedback or assistance just comment here and we can see what we can do.

arollason:
I've updated the pull request to include the no-reinit option. I removed the calls to "hasOption" because they always returned true. getOption returns the expected value of whether the flag is present in the command or not. If --reinit and --no-reinit are both used in the same command, a reinitialisation of the config cache will be run. If no flags are provided a reinitialisation will be run. If only --no-reinit is provided, no reinitialisation will be run.
Let me know if any other changes are required.

tkn98:
I'd say as reinitialisation is the default we currently have, if you specify both switches (and as we can't determine which comes last IIRC from Symfony console) --no-reinit should win if both switches are used.

Also please squash your commits into one for this feature so I can take a better view on the code at once. You can force-push the feature branch and this pull-request will automatically be updated.

arollason:
If that's the desired functionality then it probably makes sense to just remove the --reinit flag entirely, as it won't be doing anything or need to be checked in the code:

cache:flush -> Flush with a config reinit
cache:flush --reinit -> Flush with a config reinit
cache:flush --no-reinit -> Flush without config reinit
cache:flush --reinit --no-reinit -> Flush without config reinit

This way the check can simply be:

if (!$input->getOption('no-reinit')) {
  // do config cache reinit
}

And there's no reason to even include --reinit in the documentation or code, other than documenting that a config reinitialisation will occur by default when running the cache:flush and cache:clean commands. If that sounds reasonable I can update and squash the PR with the change.

tkn98:
The `--reinit` would allow to create portable scripts in the future when we change the default behavior.

arollason:
Ok no worries, I've pushed up the change that leaves the flag --reinit as an available option, and the config cache reinitialisation will occur by default. If --no-reinit is passed no reinitialisation will occur.

readme.rst
```diff
@@ -685,13 +685,17 @@ If you would like to clean only one cache type:
 
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:clean [code]
+   $ n98-magerun.phar cache:clean [code] [--reinit --no-reinit]
```
ktomk: I think this should be:

     $ n98-magerun.phar cache:clean [--reinit][--no-reinit][<code>]

This takes care about the two new options and fixes the misnomer on the operand. I'm aware this is not yet consistent style, so perhaps just for the two new options and you might want to separate by space. Please decide on your own.

readme.rst
```diff
@@ -685,13 +685,17 @@ If you would like to clean only one cache type:
...
 
 If you would like to clean multiple cache types at once:
 
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:clean [code] [code] ...
+   $ n98-magerun.phar cache:clean [code] [code] ... [--reinit --no-reinit]
```
ktomk: Similar here, please place the options before the operands.

readme.rst
```diff
@@ -685,13 +685,17 @@ If you would like to clean only one cache type:
...
+
+Options:
+    --reinit Reinitialise the config cache after cleaning (Default)
+    --no-reinit Don't reinitialise the config cache after cleaning
```
ktomk: A short notice about that --reinit overrides --no-reinit would be fine.

readme.rst
```diff
@@ -700,9 +704,15 @@ Run `cache:list` command to see all codes.
 Remove all cache entries
 """"""""""""""""""""""""
 
+Flush the entire cache.
+
 .. code-block:: sh
 
-   $ n98-magerun.phar cache:flush
+   $ n98-magerun.phar cache:flush [--reinit --no-reinit]
```
ktomk: ibid.

src/N98/Magento/Command/Cache/CleanCommand.php
```diff
@@ -29,6 +42,9 @@ protected function configure()
 
 If you would like to remove all cache entries use `cache:flush`
 
+Options:
+    --reinit Reinitialise the config cache after cleaning (Default)
+    --no-reinit Don't reinitialise the config cache after cleaning
```
ktomk: ibid.

tkn98:
Review usage manual interaction failure. The Github U/I is to be learned first. @arollason Please see the comments in the reivew, looks like I clicked once too often last night.

arollason:
Hi ktomk, thanks for taking the time to review. I've updated the documentation to reflect that --no-reinit will override --reinit. I used spaced to separate the options to be consistent with the other commands with options in the documentation already.

tkn98:
Thanks! This is now part of the development version will go into next stable.

PR #867, opened 2016-10-25
added 1.9.3.0 to config.yaml

sylink:

- [ ] Pull request against develop branch (if not, just close and create a new one against it)

Subject: Add 1.9.3.0 to installer

Fixes - Config.yaml updated to included 1.9.3.0 from the magento mirrors, handy to have.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -281,6 +281,15 @@ commands:
         extra:
           sample-data: sample-data-1.6.1.0
 
+      - name: magento-mirror-1.9.3.0
+        version: 1.9.3.0
+        dist:
+          url: https://github.com/OpenMage/magento-mirror/archive/1.9.3.0.zip
+          type: zip
+          shasum: a1f18d56c03f74207a712b12b2d93d7c524d1f1d
+        extra:
+          sample-data: sample-data-1.9.1.0
+
       - name: magento-mirror-1.9.2.4
         version: 1.9.2.4
         dist:


tkn98:
Thanks a lot for your PR. Please let me know what do you think about waiting until 1.9.3.1 comes out as we have a kind of warning to not update to 1.9.3.0? IIRC 1.9.3.1 is expected at the end of the week.

sylink:
We (we being a Magento host) use the random other versions all the time for various things so its nice to be able to call them down through n98. So I'd leave both in if possible - i.e. everything the Magento mirror repo has.

Probably it could be written to automatically know that somehow, but easier to just update a config file :)

As it stands we did this in the mean time with the ~/.n98-magerun.yaml for added commands -
`commands:
  N98\Magento\Command\Installer\InstallCommand:
    magento-packages:
      - name: 1.9.3.0
        version: 1.9.3.0
        source:
          url: https://github.com/OpenMage/magento-mirror.git
          type: git
          reference: 1.9.3.0
        extra:
          sample-data: sample-data-1.9.1.0`

tkn98:
Well actually it would be a nice idea to have some kind of provider / creator for entries based on OpenMage. So far we don't have a repository for it. But the local file as you updated it :)

joshuaswarren:
We also use random versions as well - we need both 1.9.3.0 and the upcoming 1.9.3.1 available. I actually just submitted my own pull request for this, as I didn't see this one was already open. Having this available ASAP would be great, and then add 1.9.3.1 whenever it's released.

ktomk:
We believe that the issue reported is handled in the development version. It can be upgraded to it using the `self-update` command with the `--unstable` switch.

Ref: 47de2d6ea25ec963b8d58894d508d81bd662d2e3

PR #868, opened 2016-10-26
Adding Magento 1.9.3.0 from the Magento Mirror

joshuaswarren:
Added Magento 1.9.3.0 to config.yaml and confirmed it works so the install command can now utilize 1.9.3.0.


joshuaswarren:
Just noticed this duplicates https://github.com/netz98/n98-magerun/pull/867 so closing.

PR #869, opened 2016-10-28
Optional Curl

rafaelstz:
Add curl method in README, such as magerun2.

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -50,6 +50,12 @@ Download the latest stable N98-Magerun phar-file from the file-server_:
 
    wget https://files.magerun.net/n98-magerun.phar
 
+or if you have problems with SSL certificate:
+
+.. code-block:: sh
+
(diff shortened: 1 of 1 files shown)


readme.rst
```diff
@@ -50,6 +50,12 @@ Download the latest stable N98-Magerun phar-file from the file-server_:
 
    wget https://files.magerun.net/n98-magerun.phar
 
+or if you have problems with SSL certificate:
```
tkn98: Curl is not the answer for having unspecified problems with SSL certificates, however providing the syntax is helpful. I'll take this over, just my 2 cents.

tkn98:
Thanks for taking care!

ktomk:
Ups, my mistake. This was master. Eradicating history, you'll find this in develop now. Next time please PR against develop. Thanks!

PR #870, opened 2016-11-01
Optional Curl

rafaelstz:
Optional Curl, such as https://github.com/netz98/n98-magerun/pull/869


ktomk:
Please file PRs against develop. In Develop we have the change in this case already if I'm not mistaken. It will become part of master then.

PR #874, opened 2016-11-30
Set created_at in schedule when running a cron job

ToonSpinISAAC:

- [x] Pull request against develop branch
- [x] ~README.md reflects changes~ *(no need for such a small change IMO)*

Subject: Created_at did not get set in `sys:cron:run`

- Set `created_at` in `cron_schedule` table when running a cron job. Otherwise the `created_at` field gtes set to `'0000-00-00 00:00:00'` in MySQL.

### DIFF
--- a/src/N98/Magento/Command/System/Cron/RunCommand.php
+++ b/src/N98/Magento/Command/System/Cron/RunCommand.php
@@ -84,6 +84,7 @@ HELP;
                 $schedule
                     ->setJobCode($jobCode)
                     ->setStatus(\Mage_Cron_Model_Schedule::STATUS_RUNNING)
+                    ->setCreatedAt(strftime('%Y-%m-%d %H:%M:%S', time()))
                     ->setExecutedAt(strftime('%Y-%m-%d %H:%M:%S', time()))
                     ->save();
 


ToonSpinISAAC:
I'm assuming that you will tick the checkboxes if applicable? If not I'll edit the PR. Thanks!

ktomk:
Do as you see fit, these checkboxes are merely from the PR template to remind some basics. Tick them, delete them, just ensure you've read them :)

ToonSpinISAAC:
Thanks @ktomk. Just making sure I've submitted a useful PR. Don't hesitate to let me know if you'd like something changed.

ktomk:
Thanks for providing the PR. It generally looks ok it is, there is a small thing as you do the time() formatting two times but the same, I think the use-case is good for a variable assignment. But I'll take care of that when I update the change-log.

ktomk:

ToonSpinISAAC:
Hi @ktomk, I see your point about the variable assignment. Thanks for taking care of it.

I've just tested it on my development machine and am happy to confirm that after running a `self-update` with the `--unstable` switch, the fix works as intended.

Thanks for merging! :smiley:

ktomk:
Thanks for confirmation. We're currently compiling the next stable release 1.97.23 and these fixes will be part of then. Should become available in 1-5 minutes.

PR #877, opened 2016-12-06
[fix] config.yaml - new command not added or override by previous commit

giuseppemorelli:

New `eav:attribute:create-dummy-values` and `category:create:dummy` not show in command list because config,yaml was not update or override on master branch

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -60,6 +60,7 @@ commands:
     - N98\Magento\Command\Cache\ListCommand
     - N98\Magento\Command\Cache\ReportCommand
     - N98\Magento\Command\Cache\ViewCommand
+    - N98\Magento\Command\Category\Create\DummyCommand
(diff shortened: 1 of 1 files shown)


ktomk:

@giuseppemorelli: Do you need the release as stable? If so, I'll do one. Just let me know.

giuseppemorelli:
Hi @ktomk, thanks for your quickly reply.
Actually I do not need in stable, but as you write in the changelog the commands had be in 1.97.23 version. So if someone search for that, they can't find them.
If you have time, please publish a new release or at least do an edit to your changelog in master branch. Just to be sure that people don't open issue for nothing.

ktomk:
@giuseppemorelli: Please find the commands now in stable n98-magerun version 1.97.25 :)

giuseppemorelli:
@ktomk thank you so much!

PR #882, opened 2016-12-13
Noautorehash option

### LINKED ISSUE #878: add -A option for db:console command (mysql --no-auto-rehash or -A)
If I use n98-magerun db:console, I will get a message that says:
"Reading table information for completion of table and column names
You can turn off this feature to get a quicker startup with -A"

And this will cause a significant delay before I can interact with the database (sometimes, on large DBs and servers that host many databases, it will pause for more than a minute).

This behavior also happens when you attempt to do some of the admin:user functions.

It would be great if we could have an option that would essentially send the "-A" or "--no-auto-rehash" option to/through the mysql client, so that those massive delays can be avoided.

AcousticRand:

Subject: Adding option to avoid hashing delay

Fixes #878 

- db:console now offers --no-auto-rehash option to speed up the mysql client (as if adding the -A or --no-auto-rehash option to mysql)

### DIFF
--- a/autocompletion/bash/bash_complete
+++ b/autocompletion/bash/bash_complete
@@ -156,7 +156,7 @@ _n98-magerun()
             opts="${opts} --format"
             ;;
             db:console)
-            opts="${opts} --use-mycli-instead-of-mysql"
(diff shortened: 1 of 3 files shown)


ktomk:
Thanks!

PR #894, opened 2017-02-17
Feature/install 1.9.3.2

pocallaghan:
Add OpenMage magento-mirror-1.9.3.2 to available installs.


ktomk:
@pocallaghan; Thanks for your PR. It was against the wrong branch, I took the commit into the develop branch instead.

ktomk:
We believe that the feature requested is now part of the development version. It can be upgraded to it using the `self-update` command with the `--unstable` switch.

pocallaghan:
Oops sorry about that, I did know it was supposed to be against develop. Pretty sure I rebased off of develop, so I guess I wasn't paying enough attention when making the PR. I just clicked the "Create PR" link that github displays after you push a new branch.

tkn98:
Yes, that is the problem with the default configuration here on Github. We could switch over to develop as default branch.

PR #896, opened 2017-02-28
Fix file integrity checking guide

mpchadwick:
Unless I'm missing something, the website doesn't provide an md5sum

https://files.magerun.net/

- [X] Pull request against develop branch (if not, just close and create a new one against it)
- [X] README.md reflects changes (if any)

### DIFF
--- a/readme.rst
+++ b/readme.rst
@@ -56,11 +56,11 @@ or if you have problems with SSL certificate:
 
    curl -O https://files.magerun.net/n98-magerun.phar
 
-Verify the download by comparing the MD5 checksum with the one on the website:
+Verify the download by comparing the SHA256 checksum with the one on the website:
 
 .. code-block:: sh
 
-    md5sum n98-magerun.phar
+    shasum -a256 n98-magerun.phar
 
 If it shows the same checksum as on the website, you downloaded the file successfully.
 


pocallaghan:
Seems more accurate that the old docs, but it appears `sha256sum` is not available on OS X and doesn't look like it can even be installed with brew. It can be checked using `shasum -a256 n98-magerun.phar`, might be worth mentioning.

readme.rst
```diff
@@ -56,11 +56,11 @@ or if you have problems with SSL certificate:
 
    curl -O https://files.magerun.net/n98-magerun.phar
 
-Verify the download by comparing the MD5 checksum with the one on the website:
+Verify the download by comparing the SHA256 checksum with the one on the website:
 
 .. code-block:: sh
 
-    md5sum n98-magerun.phar
+    sha256sum n98-magerun.phar
```
ktomk: Please use `shasum -a256` command instead of `sha256sum`.

ktomk:
@mpchadwick Thanks for the PR, The only thing to miss is that it recently changed away from it. As phar files can be easily manipulated, the demonstrated recent collision attack for SHA1 on PDF files should be similarly possible for PHAR files (no idea if (perhaps the compression we use makes this harder).

mpchadwick:
@ktomk @pocallaghan updated the command to `shasum -a256`

pocallaghan:
Sadly, I'm not 100% sure CentOS has `shasum`, if you wish to cover all bases I think (you|we)'ll need to include both suggestions. but not really my choice. What do you think @ktomk?

ktomk:
@pocallaghan: I'm for `shasum -a256` as per your comment on OS X. I tested it on an Ubuntu 16.04 and I remember it's available on 14.04 as well. Googling returns some users saying they don't have it on their CentOS/RHEL/Fedora boxes, but I think nowadays it should be available. If not, users should report the issue and then we'll see if it hurts anyone.

tkn98:

PR #910, opened 2017-04-19
Adds "magento" to subfolders list

mmenozzi:
Several projects we worked on had Magento into a subfolder called "magento"... So can you please add it to the subfolders list? :-)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -34,6 +34,7 @@ detect:
     - webroot
     - web-root
     - wwwroot
+    - magento
 
 event:
   subscriber:


cmuench:
Thanks for the PR. The list contains folders which are common names in web environments. The folder "magento" is not a common name.

Please feel free to add your own ~/.n98-magerun.yaml file in your home directory. The config can also defined as system wide config.
https://github.com/netz98/n98-magerun/wiki/Config

If you don't like a config you can add a "stop file" in your project root.
The file must be named ".n98-magerun" and contains the relative path to your Magento source.
I prefer this way.
See: http://magerun.net/released-n98-magerun-version-1-95-0/  "Alternative project config"

mmenozzi:
Thank you Christian! I didn't know about the stop file. It solves my problem.

Issue #806, opened 2016-04-08
Refactoring to reuse code between magerun, magerun2, conrun, ..


amenk:
* extract N98\Util to composer module

Create new composer module to comprise the reusable parts of the "xRun" tools
* extract \N98\Magento\Application\ConfigurationLoader, work with a variable config file name 

-- I might add more here --

amenk:
See also #494

ktomk:
We're currently looking to streamline code between Magerun 1 and Magerun 2 to identify parts that can be shared. This is similar to what you suggest. How this is technically done is not yet fully tested, as it's currently in the making.

amenk:
Sounds awesome! Is there a branch for this or ticket?

ktomk:
Right now I'm fixing some bugs on the road to it but it will come into develop.

M1 + M2 will also get local test setup, so it's easier to run the whole testsuite on the local box with a quick clone.

amenk:
What is the time frame for this? Anything I could help without duplicating your work? Do you make multiple composer libs?
We want to make a port for typo3 soon... Especially with the DB functionalities.

ktomk:
We currently do it this way: We duplicated the whole project and then over the time we were looking which parts are actually common. Right now some of these common parts become visible (this might be a little different to your needs, but I think you'll get a good share out of it compared to the current status quo).

For these a common location is to be established which allows to share them in a version controlled manner.

It's not necessarily as part of a composer package, but whatever we do, it must be compatible with composer. We're at the point to actually find this out, so if you know how to share a folder across multiple projects, please share your experiences:

* M1
* M2
* shared

M1+2 both use shared as a directory, it will contain a same namespace root as for example the src folder may contain as namespace in M1+2.

amenk:
I think the way to go would be to make composer libs of the components which are useful for other projects (and are shared between MR1 and MR2). Those Util libs and database functions are the most obvious for now.

ktomk:
I think this evolve from the shared folder. When it proves working, it can be turned in a git submodule and then that into a composer package. Next to these three repositories I plan to have a fourth one that is a kind of development stand to put the processes around the forks under version control. This could also help you with your plans.

So far I've extracted in the MR2 repo, a test against MR1 is pending.

Identifying and extracting the common parts is then one of the next steps. Maybe we can also meet sometime face-to-face soon again and discuss this development related topic and how to progress with it.

amenk:
[...]

PR #913, opened 2017-05-07
#806 Move all N98\Util to the shared folder, use composer (dual usage)

amenk:
The shared folder is included via composer, so it can be used as
before but also required via composer to other projects (dual usage)

- all N98\Util classes including their tests are moved to the shared
  folder
- add composer.json to the util folder, move autoload there
- require the shared folder


amenk:
Tests should work now, but build.sh still fails:

    PHP Fatal error:  Uncaught Error: Class 'N98\Util\ArrayFunctions' not found in phar:///home/self/imi/workspace/n98-magerun/build/output/n98-magerun.phar/src/N98/Magento/Application/Config.php:310
Stack trace:

no clue why ...

cmuench:
@amenk Can you change the indentation level of your comment? :-P

amenk:
done.. hopefully :-) I should have a precommit hook...

ktomk:
@amenk: git subrepo =/= composer, just in case this went under. https://github.com/ingydotnet/git-subrepo

amenk:
​Yes I know. In my PR it is still autoloaded the same way as before. The
composer.json is just "for myself" to include it via composer. It does not
interfere with the rest of magerun.

amenk:
So what's the plan?

ktomk:
From my end git subrepo.

amenk:
That you did already, right? It is still the same, the composer.json should
not hurt. Autoloading is the way you did it. I just moved all the N98\Util
to the shared folder and fixed some issues.

After you accept the PR should should be able to push to the subrepo?
​

ktomk:
Can you squash, polish and rebase the commits so that it is easier to review?

amenk:
sure, done

ktomk:
Thanks. in my opinion this is not needed and mistakes the meaning of the shared folder. I decline the PR.

You can achieve the same by using magerun as a remote for the package you would like to create and handle this via subtree merge. That will also allow you to apply patches and stay more independent.

amenk:
But you already started to move stuff to shared/ --- why not the full
N98\Util ? Which problems does it generate? I could also sync with magerun2
if you'd like...
But subtree split / merge might be an option, will think about it.

ktomk:
Yes, please checkout the subtree merging . I don't do it that often, but have this with some projects in which I need to keep track of upstream changes but I do have local changes as well.

It generates problems in that we package in here but we don't. So this contradicts. The shared folder actually is not used to extract the utils, at least not that way you do it here. So what you try to achieve with this PR might be what you want, but there is no implication for the project to have it. You need more freedom here and more control, and for contradicting things this would always cause problems.

If it's externatlized you can decide what you need for your packaging (building etc.) and what should be pushed back upstream.

Right now we already have the problem of the deviation between the mr1+2 repos. It originally was one but somehow it got cutted in half. The shared folder is a  mitigation of that while keeping the repos independent. Creating a package here wouldn't do that.

So please try if utils folder will do it for you. And sorry if that is at a different pace. From what I sense, I can imagine that it's also better for you to have more control.

amenk:
I think I would need more like to do a subtree split and then could make a composer module out of it.

     git subtree split --prefix src/N98/Util/ -b n98-util-split

Unfortunately I need two folders - also the shared one :-( So I have to use filter branch ... will see ...

amenk:
Here we go ... https://github.com/iMi-digital/n98-utils-split

PR #918, opened 2017-05-24
Use pwd to work correctly with symlinks

bramstroker:
Our magento shops use symlinks for some subdirectories to make them compatible with our deployment process. For example the media folder which contains user generated content and does not need to be overwritten after our deploy.

When we run magerun commands (for example db:dump) from this directory an exception is thrown:

```
[InvalidArgumentException]
"app/etc/local.xml"-file '/app/etc/local.xml' is not readable
```

This is because Magerun does Magento detection starting at the current working directory, which is determined by getcwd(). However this method always resolves the symlink target which obviously causes issues in our case. This change will not resolve the symlink but return the path of the symlink. I cannot think of any use case where this will cause issues for other users.

### DIFF
--- a/src/N98/Util/OperatingSystem.php
+++ b/src/N98/Util/OperatingSystem.php
@@ -113,6 +113,9 @@ class OperatingSystem
      */
     public static function getCwd()
     {
+        if (self::isLinux() || self::isMacOs()) {
+            return exec('pwd');
+        }
         return getcwd();
     }
 


ktomk:
`getcwd()` does physical resolution, `pwd` can too, so this would not be save. If you're bound to logical resolution, this would be `getenv("PWD")` (which is equivalent to `pwd --logical`).

But this would add configuration specific code which should not be necessary and prevented. For the problem you report, the *stopfile* should be the answer. Please try this first, as this keeps configuration in configuration and not within the code and also with your project.

Also please show the file-system layout with the symlinks so the concrete problem is more clear.

bramstroker:
@ktomk 

Thank you for your quick response.

Our project layout is like this:
```
/var/www/project
  public_html
    app
    skin
    var -> /var/www/project/shared/var (this is a symlink
```

When I run Magerun from /var/www/project/public_html/app everything works as expected. However when I run from /var/www/project/public_html/var it wil not find the Magento root.

I don't see why the stopfile can be a solution in our case.

ktomk:
> I don't see why the stopfile can be a solution in our case.

Placing the stopfile in `/var/www/project/` will do it for you. The feature is exactly for such cases as it provides the hint for that project where to locate the base directory of the Magento2 project so Magerun must not detect but just has it straight forward.

## Stopfile Shell Example
in your example the stopfile absolute path is `/var/www/project/.n98-magerun2` containing a line "`/var/www/project/public_html`"

~~~
$ echo "/var/www/project/public_html" > /var/www/project/.n98-magerun2
~~~

Both the logical as well as the physical paths can resolve to the stopfile which then is used to configure the base path to operate on.

ktomk:
@bramstroker: Did this solve your problem?

PR #931, opened 2017-07-21
Added 1.9.3.3 and 1.9.3.4 to the install option

michielgerritsen:

- [X] Pull request against develop branch (if not, just close and create a new one against it)
- [X] README.md reflects changes (if any)

Subject: Added missing Magento versions

I added Magento versions 1.9.3.3 and 1.9.3.4 to the install command. They where available in the "mirror" section, but not in the main section.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -178,6 +178,24 @@ commands:
 
   N98\Magento\Command\Installer\InstallCommand:
     magento-packages:
+      - name: magento-ce-1.9.3.4
+        version: 1.9.3.4
+        dist:
+          url: https://github.com/OpenMage/magento-mirror/archive/1.9.3.4.zip
+          type: zip
+          shasum: 6dadf8de337ae7033839839caeb15cbb2e98e4b8
+        extra:
+          sample-data: sample-data-1.9.2.4
+
+      - name: magento-ce-1.9.3.3
+        version: 1.9.3.3
+        dist:
+          url: https://github.com/OpenMage/magento-mirror/archive/1.9.3.3.zip
+          type: zip
+          shasum: f613968e1ea667ec1a0f0198b7c1d319e2598580
+        extra:
+          sample-data: sample-data-1.9.2.4
+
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -178,6 +178,24 @@ commands:
 
   N98\Magento\Command\Installer\InstallCommand:
     magento-packages:
+      - name: magento-ce-1.9.3.4
```
ktomk: This name is reserved for original Magento CE packages by the vendor, not OpenMage as used in this entry.

config.yaml
```diff
@@ -178,6 +178,24 @@ commands:
...
+        version: 1.9.3.4
+        dist:
+          url: https://github.com/OpenMage/magento-mirror/archive/1.9.3.4.zip
+          type: zip
+          shasum: 6dadf8de337ae7033839839caeb15cbb2e98e4b8
+        extra:
+          sample-data: sample-data-1.9.2.4
+
+      - name: magento-ce-1.9.3.3
```
ktomk: This name is reserved for original Magento CE packages by the vendor, not OpenMage as used in this entry.

ktomk:
As reported, both are already in the mirror section.

ktomk:
If you like to provide original packages, please do so. Please provide checksums and a publicly accessible download location for the packages. Please take the original packages from the vendor and document the download, there should be earlier tickets that give examples.

ktomk:
Please let me know if you think it is worth to provide original packages via a mirror.

For the flaw on the listing, the order has changed within the config file in the development version since your PR and "magento-ce" packages have been moved below all others. Please let me know if this removes the ambiguity for you or what you think.

ktomk:
@michielgerritsen  Please let me know if you think it is worth to provide original packages via a mirror.

For the flaw on the listing, the order has changed within the config file in the development version since your PR and "magento-ce" packages have been moved below all others. Please let me know if this removes the ambiguity for you or what you think.

Issue #938, opened 2017-08-30
Error after executing ./n98-magerun.phar selfupdate


tylerssn:
`./n98-magerun.phar --version`: n98-magerun version 1.98.0 by netz98 GmbH

I receive the error below. The only effort that I have made to resolve this is disabling `app/lib/n98-magerun`

```
Fatal error: Uncaught Error: Call to undefined method Symfony\Component\Console\Command\HelpCommand::isHidden() in /var/www/us/vendor/symfony/console/Descriptor/ApplicationDescription.php:122
Stack trace:
#0 /var/www/us/vendor/symfony/console/Descriptor/ApplicationDescription.php(89): Symfony\Component\Console\Descriptor\ApplicationDescription->inspectApplication()
#1 /var/www/us/vendor/symfony/console/Descriptor/TextDescriptor.php(194): Symfony\Component\Console\Descriptor\ApplicationDescription->getCommands()
#2 /var/www/us/vendor/symfony/console/Descriptor/Descriptor.php(55): Symfony\Component\Console\Descriptor\TextDescriptor->describeApplication(Object(N98\Magento\Application), Array)
#3 /var/www/us/vendor/symfony/console/Helper/DescriptorHelper.php(72): Symfony\Component\Console\Descriptor\Descriptor->describe(Object(Symfony\Component\Console\Output\ConsoleOutput), Object(N98\Magento\Application), Array)
#4 phar:///var/www/us/n98-magerun.phar/vendor/symfony/console/Command/ListCommand.php(81): Symfony\Component\Cons in /var/www/us/vendor/symfony/console/Descriptor/ApplicationDescription.php on line 122
```

JeroenBoersma:
I've the same after updating, downgraded again...

```
PHP Fatal error:  Uncaught Error: Call to undefined method N98\Magento\Application\Console\Event::setName() in phar:///usr/local/bin/n98-magerun/vendor/symfony/event-dispatcher/EventDispatcher.php:43
Stack trace:
#0 phar:///usr/local/bin/n98-magerun/src/N98/Magento/Application.php(580): Symfony\Component\EventDispatcher\EventDispatcher->dispatch('n98-magerun.app...', Object(N98\Magento\Application\Console\Event))
[...]

Issue #902, opened 2017-03-24
admin:user:unlock broken


convenient:
https://github.com/netz98/n98-magerun#unlock-admin-user
```
Unlock admin user

$ n98-magerun.phar admin:user:unlock [username]
```

I checked out the tag `git checkout tags/1.97.28` to run this test after noticing it on production.

```bash
/opt/boxen/phpenv/versions/5.5.30/bin/php /Users/lukerodgers/src/n98-magerun/bin/n98-magerun --root-dir=/Users/lukerodgers/src/project/pub admin:user:unlock -vvv
Load dist config /Users/lukerodgers/src/n98-magerun/src/N98/Magento/Application/../../../../config.yaml
Search for Magento in folder /Users/lukerodgers/src/project/pub
Found Magento in folder /Users/lukerodgers/src/project/pub
Load dist config cached
Add command  admin:notifications -> N98\Magento\Command\Admin\DisableNotificationsCommand
Add command  admin:user:change-password -> N98\Magento\Command\Admin\User\ChangePasswordCommand
Add command  admin:user:create -> N98\Magento\Command\Admin\User\CreateUserCommand
Add command  admin:user:list -> N98\Magento\Command\Admin\User\ListCommand
Add command  admin:user:delete -> N98\Magento\Command\Admin\User\DeleteUserCommand
Add command  admin:user:change-status -> N98\Magento\Command\Admin\User\ChangeStatusCommand
Add command  admin:user:lock -> N98\Magento\Command\Admin\User\LockCommand
Add command  admin:user:lockdown -> N98\Magento\Command\Admin\User\LockdownCommand
Add command  admin:user:unlock -> N98\Magento\Command\Admin\User\UnlockCommand
Add command  cache:clean -> N98\Magento\Command\Cache\CleanCommand
Add command  cache:dir:flush -> N98\Magento\Command\Cache\Dir\FlushCommand
Add command  cache:disable -> N98\Magento\Command\Cache\DisableCommand
Add command  cache:enable -> N98\Magento\Command\Cache\EnableCommand
Add command  cache:flush -> N98\Magento\Command\Cache\FlushCommand
Add command  cache:list -> N98\Magento\Command\Cache\ListCommand
Add command  cache:report -> N98\Magento\Command\Cache\ReportCommand
Add command  cache:view -> N98\Magento\Command\Cache\ViewCommand
Add command  category:create:dummy -> N98\Magento\Command\Category\Create\DummyCommand
[...]

PR #942, opened 2017-09-14
Add another fallback check to enterprise detection

convenient:

See 
- https://github.com/netz98/n98-magerun/issues/938
- https://github.com/netz98/n98-magerun/issues/902

Add an additional fallback check for the enterprise layout directory

In 902 I said 
>For visibility there is no app/code/core/Enterprise folder as Enterprise lives under the /vendor directory for this project.

There is still an Enterprise frontend directory as there is no amount of autoloading shenanigans you can do to get them to work in a modern way.

The proposed change will
- Fix enterprise detection for my use case (#902)
- Fix the broken symfony/console errors people are seeing without having to require a change in workflow (#938)


ktomk:
Thanks for the update. From my end, I'd say patch is then complete.  Thanks!

@cmuench:  Can you double check against the Enterprise file-system layout?

PR #965, opened 2017-12-24
Create new meta generator for phpstorm > 2016.1

gillg:

Update PhpStorm meta generation for newer versions

- Rename dev:ide:phpstorm:meta to dev:ide:phpstorm-old:meta
- Add new dev:ide:phpstorm-2017:meta for versions > 2016.1
- Add alias dev:ide:phpstorm:meta for current stable version


gillg:
As @ktomk suggested in !949 I set 2 long names and an alias for current version.
I'm not sure this is the best solution, because we could rename future deprecated versions in future (avoid delete for licences limited to certains versions). Another issue with alias, is autocomplete. Bash script generation not handle aliases actually.

What is your opinion ?

cmuench:
@gillg Thanks for your submission. I would prefer that the command generates the file for the latest PHPStorm version. There could be a "legacy" flag or a "phpstorm-format" option to specify the format.

Example:

`n98-magerun.phar dev:ide:phpstorm:meta --format=2016.2`

What do you think?

gillg:
I think it's a better idea. By default format should be always the last version, and if necessary user could define its own alias to not specify format each time. 
An alias of command + option could be possible ?

cmuench:
@gillg Yes, that's possible.

I have e.g. some in my local user config.

```yaml
commands:
   aliases:
     - "test-customer": "customer:create foo@example.com password123"
```

cmuench:
thx

PR #970, opened 2018-01-16
Add ability to import db from STDIN

### LINKED ISSUE #934: add --stdin option to db:import
Would it be possible to add the ability to import a DB via a pipe?

db:import always seems to expect a filename. db:dump has an --stdout parameter.

Is there a specific reason why there is no --stdin for db:import?

pocallaghan:
Subject: add --stdin option to db:import

Fixes #934 .


pocallaghan:
Not familiar enough with Travis, I don't *think* the failure is due to anything I did, but if it is, then let me know.

cmuench:
@pocallaghan There are some issues with the TravisCI. I restarted the job. Let's see how it works.

PR #993, opened 2018-07-05
Don't strip newsletter templates when stripping subscribers

scottsb:
Subject: Don't strip newsletter templates when stripping subscribers

Currently the tool strips the newsletter templates as well as the actual subscriber data (contrary to the description of the `@newsletter` and `@development` group descriptions). This avoids that.


cmuench:
Thx

PR #1005, opened 2018-10-06
add openmage-lts to installable versions

Flyingmana:
As one of the most active Magento1 Forks, we would appreciate it to also get listed as part of the installable versions.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -406,6 +406,15 @@ commands:
         extra:
           sample-data: sample-data-1.1.2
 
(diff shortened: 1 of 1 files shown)


cmuench:
Thx

sreichel:
One more reason to use N98 :)

cmuench:
@sreichel thanks. BTW it's possible to add custom import sources.
https://github.com/netz98/n98-magerun/wiki/Magento-installer#add-own-magento-repositories

PR #1019, opened 2019-05-08
Add connection option for db:dump command

### LINKED ISSUE #1018: db:dump with specified Connection
**Given** there might more than one connection in the project
**Then** it will be useful to use something like `--connection` and then pass connection name defined in **local.xml**
**Thus** allowing for db:dump to be done not from main database

**Real life scenario** given db:dump might lock database or add load to it 
**Then** a slave database is created

Ig0r-M:

Fixes #1018

- To add `--connection` option for the `db:dump` command


cmuench:
thanks

Ig0r-M:
@cmuench any idea how this can be tested though ??

cmuench:
In a Unit Test we need to mock the connection.
Most of the magerun test cases are functional tests.
For that we need a second database.

PR #1029, opened 2019-06-07
Phpstorm meta files - support dynamically called helpers + createBlock() methods

sreichel:
Phpstorm can't resolve methods if helpers are called via helper method:

```
    protected function helper($name)
    {
        return Mage::helper($name);
    }
```

```
    $this->helper('payment')->getBillingAgreementMethods()
```

This PR takes the already collected helper classes and outputs them in a new file to make them "readable" ...

```
namespace PHPSTORM_META {
    override( \Mage_Admin_Model_User::_getHelper(0),
        map( [
            'adminnotification/data' => \Mage_AdminNotification_Helper_Data::class,
           ...
```

Note: does not support old output format.

**Update:**

Detect created blocks

```
    $select = $this->getLayout()->createBlock('core/html_select')
````

### DIFF
--- a/src/N98/Magento/Command/Developer/Ide/PhpStorm/MetaCommand.php
+++ b/src/N98/Magento/Command/Developer/Ide/PhpStorm/MetaCommand.php
@@ -51,6 +51,30 @@ class MetaCommand extends AbstractMagentoCommand
     );
 
     /**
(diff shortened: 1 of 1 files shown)


cmuench:
@sreichel Thanks

Issue #1002, opened 2018-09-26
Composer conflict


kgoedecke:
Can we please update `"symfony/console": "~2.3",` to at least version 3, I'm facing conflicts because other packages depend on 3.x.

cmuench:
That is a comedy topic. We have to support multiple Magento versions at once. At the moment we are searching a solution to support the tool for Magento 2.3 where we face the same problem.

kgoedecke:
Thanks for getting back to me. Is there at least a branch which can be used with 3.x?

cmuench:
No. btw... I don't understand why so many ppl install n98-magerun with composer. One of the main ideas of the tool is that e.g. an admin can use it without any code or config change of the shop installation.
We always recommend bthe usage of the phar version.

hostep:
> I don't understand why so many ppl install n98-magerun with composer.

Because there is a `composer.json` file and the packages are published on https://packagist.org/ ? 😉 

No worries, I understand the difficulties here, using the phar file directly is a better way to manage these kind of utilities then composer, even though composer is really easy for upgrading packages and not forgetting about upgrading them.

cmuench:
@hostep Yes, as we published the composer package, we was able to support the existing Magento versions. At the moment we have trouble to support every used Symfony version in the wild.
If it is a solution, we could drop the composer support. Who can we support Symfony 2.x, 3.x, 4.x, ... components at once?

Morgon:
> btw... I don't understand why so many ppl install n98-magerun with composer. 

FWIW, I install via composer in my local environment so I can have access to it within any project, and composer 'manages' it (executable paths, updates, etc).

hostep:
It's more of a problem with composer then the magerun tool itself.
Magerun and Magento are two separate things, even though magerun can communicate with Magento, it doesn't necessarily mean that dependencies of both products should be the same or compatible for that matter.
Composer doesn't know how to deal with these kind of situations unfortunately.

There are however some composer plugins which can possibly deal with this a bit better, I tried those out about a year ago if I remember correctly, but there were some things I didn't quite like back then, maybe I should give them another try again, the ones I'm talking about are:
- https://github.com/bamarni/composer-bin-plugin
- https://github.com/phar-io/composer-plugin

Not sure if this is really helpful in this thread.

sreichel:
> No. btw... I don't understand why so many ppl install n98-magerun with composer.

Maybe because phar files aren't allowed on some shared hostings?

[...]

PR #1031, opened 2019-06-09
Update for symfony 3.x/4.x

sreichel:
See #1002

I guess it is not final (empty methods), but it works so far.

Maybe someone can test it?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,12 +18,12 @@
         "n98/junit-xml": "~1.0",
         "php": ">=5.4.0",
         "psy/psysh": "~0.4",
-        "symfony/console": "~2.3",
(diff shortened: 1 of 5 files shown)


sreichel:
Well, any ideas why tests are written to stdout and fail?

hostep:
I would use `^4.0` here as version contraint instead of `~4.0`. In this case it will act the same if you only specify a minor version and not a patch version, but it's preferred to use `^` for semver instead of `~`. Just a small remark though :)

sreichel:
Closed (for now).

Magerun suports PHP5.4, but symfony v3.2 requires PHP5.5.9 or v4 PHP7,1 ... i doubt it will be merged, if if it works. Maybe for M2 ... :)

PR #1034, opened 2019-06-16
Adding OpenMage Magento LTS 1.9.4.x

sreichel:

Add latest version to installer.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -416,6 +416,15 @@ commands:
         extra:
           sample-data: sample-data-1.1.2
 
+      - name: openmage-lts-1.9.4.x
+        version: dev-1.9.4.x
+        source:
+          url: https://github.com/OpenMage/magento-lts.git
+          type: git
+          reference: 1.9.4.x
+        extra:
+          sample-data: sample-data-1.9.2.4
+
       - name: openmage-lts-1.9.3.x
         version: dev-1.9.3.x
         source:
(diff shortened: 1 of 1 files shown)


ktomk:
@sreichel I somewhat think the number of non-versioned packages is a sanity check in the test-case. IIRC I added it really some time ago.

If there was removal of a version in this PR which still should be supported (but was removed due to the falling test as the commit msg suggests), raise the number as needed in the test.

The failing test should only remind not to add non-versions by accident (syntax of a version) or in case the version detection breaks, so self-check of the test.

sreichel:
@ktomk My guess was you don want to limit it intenionally. :)

"mageplus" is based on Magento 1.7 and not maintained anymore. Still keep it? Lts 1.9.3.x should just be removed in favour of 1.9.4.x. (1.9.3.x just gets official patches merged, all other PRs are for 1.9.4.x only)  ... Please let me know if I should re-add "mageplus".

ktomk:
If it is still available for download, I'd say it's OK to keep, e.g. if there are any users who have this in CI or similar processing. If not re-added, these would break, that is not so nice then.

Do you know it can still be installed via Magerun?

sreichel:
@ktomk  No, I dont know. Re-added and changed test.

sreichel:
OpenMage does not update 1.9.3.x branch anymore .... if you're okay with, i'd add a PR to remove it.

cmuench:
@sreichel ok

Issue #483, opened 2015-02-04
Cache clean "race" condition


electrawn:
In magerun, if I use ip addresses for the database server and change those in local.xml, when I run cache:clean config I’ll get a “no route to host” error because magerun will pull the ip address for the db server from the cache before attempting to cleaning it. 

This snippet of code below does work using array('global_ban_use_cache' => TRUE) for the init. I traced how magerun works back to the init magento1 function, however I don’t see how to easily inject global_ban_use_cache into the initsettings. This setting also only works C 1.6/E 1.11 or higher. Probably should be enabled by default for any cache clean or flush command.
```php
<?php
/**
 * Set global/skip_process_modules_updates to '1' in app/etc/local.xml and
 * then use this script to apply updates and refresh the config cache without
 * causing a stampede on the config cache.
 *
 * @author Colin Mollenhour
 */
umask(0);
ini_set('memory_limit','512M');
set_time_limit(0);
if(file_exists('app/Mage.php')) require 'app/Mage.php';
else require '../../app/Mage.php';

// Init without cache so we get a fresh version
Mage::app('admin','store', array('global_ban_use_cache' => TRUE));

// Now enable caching and save
Mage::getConfig()->getOptions()->setData('global_ban_use_cache', FALSE);
Mage::app()->baseInit(array()); // Re-init cache
Mage::getConfig()->loadModules()->loadDb()->saveCache();
echo "Saved config cache.\n";
```

cmuench:
@electrawn 
We fixed that in the latest 1.95.0 release. Can you verify it?

amenk:
this causes a regression: #613 This ticket seems like quite an edge case. What about adding a switch `--reinit` to the `cache:flush` for such cases?

ktomk:
Original source of the script presented here in OP is: https://gist.github.com/colinmollenhour/2715268

IMHO the race-condition reported here is no race condition, it's a configuration issue (or question).

The reported problem (“no route to host” error) is a Magento misconfiguration. The database is not available any longer. The problem is that the database configuration has been changed without clearing / disabling the config cache first. It then might be too late (as in the given scenario) because the former database configuration is not accessible any longer and Magerun quits with an error.

The changes applied should be reverted, Magerun can only relate to the concrete configuration which might - or might not - be cached. It's up to the user to configure the system properly which includes cache invalidation, Magerun (still) can't do that for the user.

[...]

PR #1041, opened 2019-08-26
Change config init to Magento's reinit function

### LINKED ISSUE #583: Redis and reached 100 router match iterations problem after cache:clean
Did anyone find a problem on running cache:clean with some option, e.g:
n98-magerun.phar cache:clean block_html
n98-magerun.phar cache:clean eav layout
with Redis and Cm_Cache_Backend_Redis module? After that kind of operation I got " Front controller reached 100 router match iterations". It looks like main routers (Admin, Standard) are not on the list anymore. What's more curious when I run
n98-magerun.phar cache:clean
...it works and I can get my store without any error anymore.

I think it could be connected with AbstractCacheCommand.php->reinitCache()
\Mage::getConfig()->loadModules()->loadDb()->saveCache();

DuckThom:
Applies the change suggested in the comments of issue https://github.com/netz98/n98-magerun/issues/583

Subject:

Fixes #583 

- Changes the config init logic to Magento's own reinit function in the Config model to prevent issues with the `CONFIG` cache tag being cleared without being rebuilt when using the Redis cache storage engine.

### DIFF
--- a/src/N98/Magento/Command/Cache/AbstractCacheCommand.php
+++ b/src/N98/Magento/Command/Cache/AbstractCacheCommand.php
@@ -83,8 +83,7 @@ class AbstractCacheCommand extends AbstractMagentoCommand
         }
 
         \Mage::getConfig()->getOptions()->setData('global_ban_use_cache', false);
-        \Mage::app()->baseInit(array()); // Re-init cache
-        \Mage::getConfig()->loadModules()->loadDb()->saveCache();
+        \Mage::getConfig()->reinit();
     }
 
     /**


cmuench:
@DuckThom Thanks for your PR. Your commit reverts some lines of code which were introduced to prevent a race condition.
We added that code in 2015.
https://github.com/netz98/n98-magerun/issues/483

We have to check if that really fixes the problem without creating a regression bug.

DuckThom:
@cmuench I see, I tested this change on my local environment yesterday and I did not find any issues running the `cache:clean` command after the change and the issues I was having were fixed.

While debugging my issue yesterday I came across this repo [AmpersandHQ/magento-ce-ee-config-corruption-bug](https://github.com/AmpersandHQ/magento-ce-ee-config-corruption-bug ) which describes the bug the lines removed in this PR were causing. 

Magento created a fix for it, but these removed lines were bypassing that fix. By calling `reinit` on the Config model it will match Magento's cache reinitialisation logic and include the bugfix. (See the latest 2 comments on 583 by convenient and mseliger-dotsource). Or do you mean a different race condition?

cmuench:
@DuckThom I had never had this issue with the race condition. As I imagine was the problem in combination with Redis and a huge amount of data.
For me it's getting harder and harder to track all the issue because the most of our projects are running on Magento 2.
In Magento 2 I can see some changes according to cache locking issues.
In which release was can I find this cache fix you mentioned?

DuckThom:
@cmuench Me neither, the project I was working on was the first one I saw having this issue as well and it's quite a large project so there is a lot of config data involved.

From what I've read, the fix was released in 1.9.3.1.

PR #891, opened 2017-02-01
Add enterprise mview reindex commands

convenient:
I've spent too long in my time as a Magento developer teaching the members on my team that the full flat indexer, and the changlog (mview) reindexer do vastly different things under the hood. It's fine to have your product reappear on the frontend when running a full reindex, but when you have to debug and fix something in the changelog process you need a finer level of granularity. Having to debug these on local, production, and every environment in between has left me wanting a standard command to handle this.

I've added two.

1. Show a list of the mview metadata, including how many entries are pending a reindex (can be useful if an indexer gets stuck)

```bash
$ magerun index:list:mview

+---------------------------------+------------------------------------+------------------------------------+--------+------------+-------------------------+
| table_name                      | view_name                          | changelog_name                     | status | version_id | entries pending reindex |
+---------------------------------+------------------------------------+------------------------------------+--------+------------+-------------------------+
| cataloginventory_stock_status   | cataloginventory_stock_status_view | cataloginventory_stock_status_cl   | 1      | 5262044    | 0                       |
| enterprise_url_rewrite_redirect | enterprise_url_rewrite_redirect    | enterprise_url_rewrite_redirect_cl | 1      | 4117       | 0                       |
| enterprise_url_rewrite_category | enterprise_url_rewrite_category    | enterprise_url_rewrite_category_cl | 1      | 73207      | 0                       |
| enterprise_url_rewrite_product  | enterprise_url_rewrite_product     | enterprise_url_rewrite_product_cl  | 1      | 29065      | 1                       |
| catalog_category_product_index  | catalog_category_product_view      | catalog_category_product_index_cl  | 1      | 455375     | 0                       |
| catalog_category_product_cat    | catalog_category_product_cat_view  | catalog_category_product_cat_cl    | 1      | 78188      | 0                       |
| catalog_product_index_price     | catalog_product_index_price_view   | catalog_product_index_price_cl     | 1      | 12189944   | 0                       |
| catalog_category_flat           | catalog_category_view              | catalog_category_flat_cl           | 1      | 107413     | 0                       |
| catalog_product_flat            | catalog_product_view               | catalog_product_flat_cl            | 2      | 0          | 0                       |
+---------------------------------+------------------------------------+------------------------------------+--------+------------+-------------------------+
```

2. Trigger a  mview changelog reindex for a given table

```
$ magerun index:reindex:mview enterprise_url_rewrite_product
[...]

PR #1044, opened 2019-12-16
Fix instances of PHP Fatal error:  Uncaught ArgumentCountError

convenient:
```
PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function N98\Magento\Command\AbstractMagentoCommand::_getModel()
```

When I built these commands in https://github.com/netz98/n98-magerun/pull/891 I erroneously neglected to pass the 2nd argument to `_getModel` and `_getSingleton`

I was using this today on a client running PHP 7 and this is now a proper error, previously I imagine a warning or notice slipped by on PHP 5.


cmuench:
Thanks @convenient

Issue #1127, opened 2020-11-27
mysqldump: Error: 'Access denied; you need (at least one of) the PROCESS privilege(s) for this operation' when trying to dump tablespaces


empiricompany:
default db:dump on centos 8 / percona server 5.7 return this error:
mysqldump: Error: 'Access denied; you need (at least one of) the PROCESS privilege(s) for this operation' when trying to dump tablespaces

i have to add "process" privilage to mysql user or pass extra option --no-tablespaces to mysqldump to avoid the error 
db:dump --dump-option="--no-tablespaces"

we can add --no-tablespaces option by default in mysqldump command?

nice explatation of tablespaces
https://anothercoffee.net/how-to-fix-the-mysqldump-access-denied-process-privilege-error/

cmuench:
@empiricompany for n98-magerun2 we have already a solution for that issue.
We need to backport that feature.

empiricompany:
@cmuench i created a new pr #1132 with the backport of no-tablespaces option from magerun2
I would have preferred this to be the default but I followed the magerun2 way

PR #1132, opened 2020-11-30
add "no-tablespaces" option

empiricompany:
Issues #1127:

backport --no-tablespaces option from n98-magerun2


cmuench:
@empiricompany thanks. Seems that Travis upgraded the default composer version. So.. no worry about the broken build. We will fix that, soon.

Issue #1149, opened 2020-12-24
Unshare Shared


ktomk:
Looks to me the days of shared are over hence I propose to unshare the files in `shared` for Magerun.

I'm willing to do this, but would be good to know if you see anything against it @cmuench 

Follow-Ups:
- [x] Confirmation from Christans drone server for project phar builds
- [x] Check PSR-4 for project autoloading (filed per #1151)
- [x] Care-taking for magerun 2 to streamline unsharing + findings (filed in https://github.com/netz98/n98-magerun2/issues/693)
- [x] Check Required minimum PHP Version (testing w/ project PHPUnit requires PHP 7.1 something [via composer system-reqs checks])

cmuench:
@ktomk I already removed the shared directory in n98-magerun2.
I think it's ok to unshare also for n98-magerun1. It's not providing too much synergies anymore.

ktomk:
Those were the days ... ACK its more old cruft, lets clean things up here.

cmuench:
Build and deploy of the phar to files.magerun.net was successful. No problem at all.

ktomk:
Declare victory.

cmuench:
@ktomk Thanks for sharing the info about the unshare shared victory.  😅

PR #1150, opened 2020-12-24
Unshare Shared

ktomk:
Since 43933ce a shared directory between Magerun 1 & 2 was in use but over
time went out of business as Magerun 2 is not participating any longer and
it has been removed there.

Remove the "shared" folder in Magerun 1 to finish the life-cycle as it has
no benefit any longer and only stands in the way maintaining the project.

The php_cs configuration has been overridden since f31c4ac in project,
delete shared one - new one is green for old shared files.

NOTE: PHP 5.3 backwards compatible code might be left in
      ArrayFunctions.php - PHP 5.3 may not be needed any longer for
      Magerun 1 (hopefully). Out of scope for this change-set thought.

Ref: 43933ce52e58637fca98a25f21a50eda06a36378 (Birth)
Ref-Internal: 472e26e8f4a69cb7b3e85745ee0e2a3963682447
Ref: f31c4acf491a7a1fe40ece2d13ac19c41bdb6c2a (PHP-CS-Fixer config)
Related-Issue: #1149

---


ktomk:
@cmuench: please review:

this was merely straight forward and github CI verifies it is working
(build and phpunit were affected).

there was also a .php_cs on it's own, I identified it as the old version
from the project (which makes sense for the original share), it has been
superseded some time ago so I dropped it.

just asking you take a look with a pair of fresh eyes.

and: if there are off-project files related to the build process (e.g.
the phar file online), I don't see them and can't speak for them, so in
case there is something, please give a short notice.

composer.json
```diff
@@ -39,7 +39,6 @@
             "N98": "src"
         },
         "psr-4": {
```
cmuench, line 41: Can be removed?

cmuench (commented):
Looks good so far. We can merge it to develop then I can check the build process which is currently outsourced to my local drone server.

composer.json
ktomk: Yes, can be removed. Left it in by intention as a reminder to check if the project can't go to PSR-4 and then would reduce diff display. I would check that in a follow-up and leave it like this in this PR. OK?

ktomk:
Okay, let's do this. Merged.

PR #1151, opened 2020-12-24
Change autoloading from PSR-0 to PSR-4

ktomk:
Previously while unsharing the shared folder in #1149 it came to attention 
that the autoloading in projects composer.json was still configured as
PSR-0 which is deprecated since longer.

Change is to switch to PSR-4 autoloading.

- [x] update for n98/junit-xml 1.1.0 (affects internal test)
- [x] check if the PSR-4 works also in combination with n98-magerun modules

---

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,15 +35,13 @@
         "mikey179/vfsstream": "^1.6"
     },
     "autoload": {
-        "psr-0": {
-            "N98": "src"
-        },
         "psr-4": {
+            "N98\\": "src/N98"
         }
     },
     "autoload-dev": {
-        "psr-0": {
-            "N98": "tests"
+        "psr-4": {
+            "N98\\": "tests/N98"
         }
     },
     "authors": [
--- a/composer.lock
+++ b/composer.lock
@@ -529,25 +529,25 @@
         },
         {
             "name": "n98/junit-xml",
-            "version": "1.0.0",
+            "version": "1.1.0",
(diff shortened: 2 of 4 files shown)


ktomk:
Straight forward one again, only little change in the prefix, the one internal test that checked it previously has been extended for PSR-4 (for PSR-0 it reveals now the `N98\JUnitXml` package prefix to keep it).

Should work as CI shows, magerun.phar build + smoke tested locally in isolation. LGTM from my side if you don't spot anything @cmuench

cmuench:
@ktomk thanks. I will check if the PSR-4 works also in combination with n98-magerun modules.
btw... I just upgraded the `n98/junit-xml` package to 1.1.0 which comes with a PSR-4 autoloader.

ktomk:
Alright, make sense with the module check, I did not look into that part. Let me know if I can do anything there.

I added the update of n98/junit-xml to 1.1.0 and changed the internal test accordingly.

Main change is still on last commit on top.

cmuench:
@ktomk Checked the modules. Everything is ok.

cmuench (approved):

ktomk:
Yes, merged. Also did some testing with modules and could not bring it to halt. It is even possible to register in the same PSR-4 prefix `N98\` for a module, e.g. `N98\FooCommand` did work for me from a module/plugin folder.

PR #1146, opened 2020-12-22
Start migration from Travis-CI to Github CI

ktomk:
Porting five build jobs from Travis-CI to Github Actions across two PHP versions and a Mysql service.
    
Straight forward port from .travis.yml with the help of the shivammathur/setup-php and ktomk/run-travis-yml Github Actions.
    
Ref-Issue: #1138

---

Magerun pull-request check-list:

- [X] Pull request against develop branch (if not, just close and create a new one against it)

Issue #1138, opened 2020-12-08
Migrate to Github Actions


ktomk:
Raised in #1133 by @cmuench, having an issue of it's own.

I ~~have plans to~~ do this as I currently already migrate in another project.

* [x] port/run Travis-CI build-jobs as Github Actions
* [x] reduce Travis-CI builds to `master` to verify migration baseline for some time
* [x] disable pull-request builds on travis-ci.org repository settings ([ref](https://travis-ci.org/github/netz98/n98-magerun/settings)).
* [x] re-integrate (parts of) [boilerplate workflow](https://github.com/actions/starter-workflows/blob/main/ci/php.yml)
* [x] use `$default-branch` macro
* [x] fix [`composer validate`](https://getcomposer.org/doc/03-cli.md#validate) (was not marking the build job red on validation warnings)
* [x] add test-setup job
* [x] streamline build files/scripts from `shared` folder (via #1149)
* [x] sunset circleci as not in use any longer (via #1152)
* [x] fix to build on master (missed in v2.1.0 - fixed but needs new release to confirm)
* [x] Sunset Circleci (via #1152)
* [x] streamline other build scripts (follow-up in part?) (via #1273)

Follow-Up Actions / Findings:
* [x] Restore Codecov (deactivated ca. two years ago) (via #1273)
* [x] The projects Composer validation is w/o dependencies (it fails, added with allowed failure to *PHP Composer* job) ([ref](https://github.com/netz98/n98-magerun/runs/1595449597?check_suite_focus=true#step:9:5), [ref](https://github.com/netz98/n98-magerun/runs/1595449597?check_suite_focus=true#step:9:16), [ref](https://github.com/netz98/n98-magerun/runs/1595449597?check_suite_focus=true#step:9:79)). Despite the warning, this does not (yet?) fail with composer v2 (at least not as dependency as this build job is using composer 2.0.8).

cmuench:
For documentation... we have already this for n98-magerun2: https://github.com/netz98/n98-magerun2/tree/develop/.github/workflows

PR #1152, opened 2020-12-25
Sunset Circleci

ktomk:
This change-set effectively migrates what has been run on Circleci way
back in the past (~ two years ago) to Github CI. And in short resurrects
code-coverage (and Codecov).

Circleci was in use to run Magerun from source to setup a Magento version
with sample data, run some smoke-tests, then run the Phpunit test-suite
with code-coverage and uploading the CLOVER_XML to the Codecov service.

Circleci eventually spit errors from which it could not recover and then
the build was (remotely) disabled. [0]

In this change-set:

- PHP version 5.5 to 7.3 (pinned 5.5.21 to unpinned 7.3.x/24)
- Magento version 1.9.2.1 to 1.9.4.5 (via MAGENTO_VERSION)
- Upload to Codecov with Github-Action codecov/codecov-action@v1
- Remove of circle.yml, the Circleci configuration file
- Add SETUP_DB_PASS to pass mysql password (as in local/test_setup.sh).
- phpenv was in use at Circleci to configure PHP and is removed as on
  Github CI the settings are already fine: Xdebug is configured for
  coverage and memory_limit=512M was required initially on Circleci to
  have enough memory for code-coverage not necessary on new runner.
- Circleci specific environment parameters have been removed:
  CIRCLE_ARTIFACTS and CIRCLE_TEST_REPORTS. The later appeared already
  unused when uploading while porting. The fall-back directory was ".",
  change is to "./build/coverage".

Notably not changed:

- If code-coverage upload to Codecov fails, the build is painted green
  despite the error. This is by intention (and can be changed with
  upload parameters).

Fun-Fact: Using the orignal token from codecov.yml while testing, the
  codecov github action detected the project as netz98/n98-magerun2 - not
  n98-magerun.

Related-Issue: #1138
[0]: https://github.com/netz98/n98-magerun/pull/1146#issuecomment-749714114

---


ktomk:
After starting the migration to Github Actions/CI (#1146) and migrating the ground-level test-setup to it (#1147) a couple of build scripts have been identified to (longer time disabled) Circleci (dedicated to keep track of code-coverage metrics (published via Codecov) over time).

Regardless of first-hand it is wanted to resurrect the use of Codecov itself (which does not look like a bad service to me), migrating what was running on Circleci to Github CI reveals some benefits like refreshing the scripts (incl. removing files) and verifying that Codecov would still work with the project.

In short this change-set carries the trash out the earlier Circleci error has left.

@cmuench I would greatly appreciate it to remove this old cruft while maintaining Codecov as a test here. There is one parameter for the coverage percentage at 66% which would make the build fail if coverage goes below, currently there is more than ~69% percent covered so I would say this is practically safe (switching to xdebug 3 should not make it fail and also for normal PRs while maintaining Magerun 1 - and even if the build reds with ccov, I'd like to see it first before lowering the number (maybe again with a 3% margin) so to know the limit still is in effect and checks are working.

On the technical side Codecov says that with the (here introducing)  Github Action no upload token configuration is necessary. I have tested this in my fork and can confirm it so far, and it appeared to me kinda practical. No idea on the security implications this has, in any-case, the token in codecov.yml (deletion in PR) did show as _Magerun 2_ with the action. So all-in-all this change looks beneficial to me also for maintaining all these services in general.

After done & in master, IMHO you could consider to switch off circle-ci for Magerun 1 completely. That is b/c from what I remember Circleci was in use for Codecov only.

cmuench (approved):
@ktomk I already dropped access of circle-ci to our repos. So we can moved forward.

ktomk:
Haha, just messed it up with the last change of the path name. I'll fix it and merge to develop.

ktomk:
Fixed + Wow Codecov really shines after these more than two years. Theres a bit to catch up, already reported it and response is looking good to remove some noise from the Github Action logs (ETA next week, patches are already proposed). Maybe there is some more for the CI integration they have but the current results are just steady.

PR #1243, opened 2022-04-27
chore: Dependabot check Github Actions daily

neilnaveen:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,4 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
-
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


ktomk:
Thanks for taking care @neilnaveen, could you please take a look at the commit message:

- limit the line length to 72 characters in the message, if you have links, reference them and add them to the end of the message, like where the sign-off line was placed.
- for the sign-off please use a real name and a real mail address. if you don't like that, just leave it out (we don't require sign-offs)
- for the subject line, please keep it short, e.g. "chore: Dependabot check Github Actions daily" or similar and restrain from past-tense as in "Included".

When you address any of these changes, please amend the commit and force-push to update the PR (instead of closing and creating a new one).

Thanks a lot!

PR #1281, opened 2023-05-22
Fixes the "always" cron expression check

### LINKED ISSUE #1280: TypeError when running sys:cron:list on PHP 8
## Describe the bug

With PHP 8.0 (and, I suspect, all later versions), the `sys:cron:list` command throws
```Uncaught TypeError: array_slice(): Argument #1 ($array) must be of type array, null given in phar:///usr/local/bin/n98-magerun.phar/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php:103```

## Expected behaviour

Launching `sys:cron:list` I expect to see a list of all available crons.

## Steps to reproduce the issue

Install OpenMage 19.4.19 on PHP 8.0 (or later PHP versions)
Launch n98 sys:cron:list

## Technical details

- Host Machine OS: Linux
- PHP version: 8.0
- Magento version: OpenMage 19.4.19

## Possible Fix

There are two issues here.
The first is that the [check for the "always" cron expr](https://github.com/netz98/n98-magerun/blob/develop/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php#L88) is done incorrectly. It assumes that the provided cron expr is a string and, as such, uses the === operator for the comparison. In reality the expression is an instance of `Mage_Core_Model_Config_Element`, so the strict comparison always fails.
When this happens, the $array variable on line 101 is set to `null`, which leads to the second issue.
Up to PHP 7.4 the `array_slice` function had no argument type specified. It accepted `null` as the first argument and carried on, merely emitting a warning. Starting with PHP 8 the type of the first argument is specified as `array`, hence the error when the function is called with `null`.

I made a PR to fix the "always" expression check. This should fix the issue and prevent `null` values from being supplied to the `array_slice` function.

azambon:

- [ ] Pull request against develop branch (if not, just close and create a new one against it)

Fixes #1280 

The cron expression that is passed to parseCronExpression() is not a string but a Mage_Core_Model_Config_Element instance.
Of course comparing a string to an object with a strict check will always return false. This PR explicitly casts the value to `string`, which has the effect of invoking the object's `__toString` method.

### DIFF
--- a/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php
+++ b/src/N98/Magento/Command/System/Cron/AbstractCronCommand.php
@@ -85,7 +85,7 @@ abstract class AbstractCronCommand extends AbstractMagentoCommand
      */
     private function parseCronExpression($expr)
     {
-        if ($expr === 'always') {
+        if ((string)$expr === 'always') {
             return array_fill(0, 5, '*');
         }
 


cmuench:
@azambon Thanks. I will review this PR soon.

cmuench:
@azambon looks good. Merged. Ignore the CI setup here. It's currently broken. We handle that in a different branch.

azambon:
Thank you @cmuench :+1:

PR #1433, opened 2024-06-11
Enable funding?

sreichel:
@cmuench just an idea to help me supporting M1-version. Feel free to close when you disagree with it.


cmuench:
@sreichel Until today we did not have any monetary funding. I am not against that but let's have a talk (maybe remote) about the maintaining and such topics.

sreichel:
Thanks. Lets talk next week. :+1:

PR #1455, opened 2024-07-31
Added php-cs-fixer to workflow

sreichel:
Add basic config and added to github workflow


sreichel:
@cmuench thanks for write permissions.

But i'll still wait for you. 

Maybe there are some changes that can be ported to M2-version too. So i feel better if you take a look at it :)

PR #1653, opened 2026-04-16
Fix strict_types errors in DumpCommand and extension autoloader registration

ajbonner:
This PR fixes two issues I ran into after building from develop:

  - Fix extension autoloader registration: Extensions declaring autoloaders or autoloaders_psr4 in their
  n98-magerun.yaml were silently ignored because the Composer classloader was running in authoritative classmap
  mode, which bypasses PSR-0/PSR-4 lookups entirely. The classloader now disables authoritative mode when an
  extension needs to register custom autoloaders.

  - Fix DumpCommand strict_types error: The $tableDefinitions property was uninitialized and guarded by an
  is_null check that fails under strict_types. Removed the unnecessary memoization ($tableDefinitions,
  $commandConfig) and the deprecated getTableDefinitions() method, inlining the single call directly via the
  database helper.

### DIFF
--- a/src/N98/Magento/Application/Config.php
+++ b/src/N98/Magento/Application/Config.php
@@ -185,6 +185,12 @@ class Config
     {
         $mask = '<debug>Registered %s autoloader </debug> <info>%s</info> -> <comment>%s</comment>';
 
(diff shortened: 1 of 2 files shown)


sreichel:
Thanks. Cli support will move to Openmage soon.

